PR #1, opened 2025-07-07
add superstructure and subsystems

spellingcat:


src/main/java/frc/robot/Superstructure.java
```diff
@@ -0,0 +1,104 @@
...
+      IntakeSubsystem intake
+  ) {
+    this.elevator = elevator;
+    this.shoulder = shoulder;
+    this.intake = intake;
+
+    addTransitions();
+  }
+
+  public void periodic() {
```
kevinclark, line 162: This smells funny but maybe I'm misreading the docs on [triggers](https://docs.wpilib.org/en/stable/docs/software/commandbased/binding-commands-to-triggers.html). Can you tell me more about what your intention is here?

src/main/java/frc/robot/Superstructure.java
```diff
@@ -0,0 +1,104 @@
...
+        return;
+      }
+    }
+  }
+
+  public void addTransitions() {
+    //TODO
+  }
+
+  private Command forceState(SuperState nextState) {
```
kevinclark: Nitpick: Method make sense but name seems kind of weird. Is there some kind of transition that's passive and not forced that this is in contrast to? I'd consider `transitionTo`, `moveTo`, `goTo`, `become`.

src/main/java/frc/robot/Superstructure.java
```diff
@@ -0,0 +1,104 @@
...
+              this.state = nextState;
+              setSubstates();
+            })
+        .ignoringDisable(true);
+  }
+
+  private void setSubstates() {
+      elevator.setState(state.elevatorState);
+      shoulder.setState(state.shoulderState);
+      intake.setState(state.intakeState);
```
kevinclark: Are there situations where these state transitions need to be coordinated or is it enough that each element knows its role? Do you have to worry about collisions or other conflicts?

src/main/java/frc/robot/elevator/ElevatorSubsystem.java
```diff
@@ -0,0 +1,49 @@
...
+
+  //TODO: add IO layers
+  /** Creates a new ElevatorSubsystem. */
+  public ElevatorSubsystem() {}
+
+  @Override
+  public void periodic() {
+    // Sets the elevator to the height corresponding to the current state of the robot
+    // Shoulder, intake, and any future subsystems should more or less follow this same pattern (additional logic might be needed though)
+    setExtension(() -> state.extensionMeters);
```
kevinclark: If the elevator's concept of state is really just a distance, is there a reason for it to wait to go to that distance until it hits periodic and should we be repeatedly calling for that extension to happen? Is this effectively a way of repeatedly adjusting PID? Tell me about your plans here.

kevinclark (commented):
I think I mostly see where you're going and it's *hand waves* in a reasonable direction. There's some choices that seem a little suspect I'd like to hear more about.

src/main/java/frc/robot/Superstructure.java
spellingcat: This should check to see if there's a transition that's being requested (currently in the start state + its corresponding trigger is true) and then if there's a command that goes with it (like autoalign, funnel etc) run that. I guess I wouldn't need to check the triggers every loop?

src/main/java/frc/robot/Superstructure.java
spellingcat: this was a holdover from inseason

src/main/java/frc/robot/Superstructure.java
spellingcat: The idea is that there are a lot of explicit intermediate states that generally don't move multiple mechanisms at once and instead splits those movements up into distinct transitions, so as long as it goes to all the correct intermediate poses it should not collide. I also put together [a doc](https://docs.google.com/document/d/1z6lj8qSADNNd7unU1ZCrEWAM7g3oJyfvDSwORe1Tlco/edit?tab=t.0) for reference

src/main/java/frc/robot/elevator/ElevatorSubsystem.java
spellingcat: I'm not quite sure what you mean, sorry- do you mean I should bind this to a trigger too?

src/main/java/frc/robot/Superstructure.java
kevinclark: I expectation is that whenever you set that there needs to be a new state, you have an opportunity to trigger the transition. Unless there's some coordination needed (f.e if two different incompatible actions could be taken and you needed to wait for those things to be triggered and then decided on) I would assume this is an on-demand action, not a polling action. No reason to pay the price every loop.

src/main/java/frc/robot/Superstructure.java
kevinclark: Release yourself from the shackles of the past ;)

src/main/java/frc/robot/Superstructure.java
kevinclark: Got it. So long as there is *never* coordination required, this seems fine. It *does* mean that you're explicitly coupling a group of states together. f.e: every time the elevator changes state, you also have to specify the shoulder and intake state. Maybe that's fine and maybe that's awkward. Keep an eye out for the latter.

src/main/java/frc/robot/elevator/ElevatorSubsystem.java
kevinclark: If you put something in periodic, it's going to be called every loop. Right now, setExtension is a Command that gets executed. Do you mean for that to happen every loop? Would it make more sense for this to just be the default command of the subsystem and you instead handle this behavior via assigning a setpoint?

spellingcat: ah, I forgot default commands existed

src/main/java/frc/robot/Robot.java
```diff
@@ -177,6 +178,7 @@ public Robot() {
     SmartDashboard.putData("l1", superstructure.changeStateTo(() -> SuperState.L1));
     SmartDashboard.putData("l2", superstructure.changeStateTo(() -> SuperState.L2));
 
+    // surely there's a better way to do this but i'm sort of a clown and also lazy
```
kevinclark, line 193: No better time to dig a little deeper than the offseason ;) What don't you like here and what do you imagine might make it better?

src/main/java/frc/robot/Superstructure.java
```diff
@@ -154,47 +155,12 @@ public Superstructure(
...
     addTransitions();
-    // addSimTransitions();
   }
 
   public void periodic() {
     Logger.recordOutput("Superstructure/Superstructure State", state);
   }
 
-  // for testing only!!
-  public void addSimTransitions() {
```
kevinclark: No longer needed?

src/main/java/frc/robot/Superstructure.java
```diff
@@ -225,10 +191,18 @@ public void addTransitions() {
     transitions.add(
         new Transition(
             SuperState.READY_CORAL,
+            SuperState.PRE_L1,
+            new Trigger(() -> Robot.getCurrentCoralTarget() == ReefTarget.L1)
+                .and(new Trigger(() -> arm.hasCoral()))
```
kevinclark: Nitpick and non-essential, but `new Trigger(arm::hasCoral)` reads moderately cleaner imo.

src/main/java/frc/robot/Superstructure.java
```diff
@@ -225,10 +191,18 @@ public void addTransitions() {
...
+                .and(Robot.preScoreReq)));
+
+    transitions.add(
+        new Transition(
+            SuperState.PRE_L1,
             SuperState.L1,
             new Trigger(() -> Robot.getCurrentCoralTarget() == ReefTarget.L1)
-                .and(Robot.scoreReq)
-                .and(new Trigger(() -> arm.hasCoral()))));
+                .and(new Trigger(() -> arm.hasCoral()))
```
kevinclark: Here too.

src/main/java/frc/robot/Superstructure.java
```diff
@@ -314,6 +288,38 @@ public void addTransitions() {
...
+            SuperState.READY_CORAL,
+            SuperState.PRE_L4,
+            new Trigger(() -> Robot.getCurrentCoralTarget() == ReefTarget.L4)
+                .and(Robot.preScoreReq)));
+
+    transitions.add(
+        new Transition(
+            SuperState.PRE_L4,
+            SuperState.L4,
+            new Trigger(() -> this.atExtension()).and(Robot.scoreReq)));
```
kevinclark: Similarly: `new Trigger(this::atExtension).and(Robot.scoreReq)`

src/main/java/frc/robot/Superstructure.java
```diff
@@ -314,6 +288,38 @@ public void addTransitions() {
...
+
+    transitions.add(
+        new Transition(
+            SuperState.POST_L4,
+            SuperState.IDLE,
+            new Trigger(
+                () -> !arm.hasCoral() && this.atExtension() && !Robot.scoreReq.getAsBoolean())));
+
     // maps triggers to the transitions
     for (Transition t : transitions) {
```
kevinclark: Is there a reason to store the transitions in the class and only bolt state transitions on to their triggers at the and of addTransitions? I.e does the `transitions` variable have a reason to exist? Would this goal be served just as well if you had a helper `` that has a body:

```java
private void bindTransition(SuperState start, SuperState end, Trigger trigger) {
    trigger.and(new Trigger(() -> state == start)).onTrue(changeStateTo(end));
}
``` 

and then you replace this:

```java
    transitions.add(
        new Transition(
            SuperState.IDLE,
            SuperState.PRE_INTAKE_CORAL_GROUND,
            Robot.intakeCoralReq.and(() -> !arm.hasCoral())));
```

with this:

```java
    bindTransition(
            SuperState.IDLE,
            SuperState.PRE_INTAKE_CORAL_GROUND,
            Robot.intakeCoralReq.and(() -> !arm.hasCoral()));
```

And delete that loop over `transitions`, the `Transition` record, and the variable entirely?

kevinclark (commented):
I'm assuming the magic numbers on the Arm make sense and you've tested them in sim?

Had a suggestion that might smooth out your state transition definitions. Non blocking, but worth considering maybe.

spellingcat:
> I'm assuming the magic numbers on the Arm make sense and you've tested them in sim?

oh not at all. I took them from kelpie and I'm not convinced that my sim setup isn't completely wrong (50 is a ridiculous number for maxvel/accel but it seemed to make sim run??) but I'll get to tuning that sort of thing next, once I've confirmed that the state transitions work

src/main/java/frc/robot/Robot.java
spellingcat: This feels like an awful way to pass data back and forth between SmartDashboard and code for tuning things- these particular lines will get replaced as soon as I add in the controllers sim but for things like real life PID tuning I'll definitely make something better. 581's DogLog and 6328 both have some kind of tunable value, which are on my todo list of things to adapt later in the offseason

src/main/java/frc/robot/Superstructure.java
spellingcat: this was from when I was trying to test the state transitions one by one for the first time

src/main/java/frc/robot/Superstructure.java
spellingcat: can you use this syntax when negating something? (eg `() -> !arm.hasPiece()`)

src/main/java/frc/robot/Superstructure.java
spellingcat: that definitely makes sense, thank you!

src/main/java/frc/robot/Superstructure.java
kevinclark: Yep, just use `Predicate.not`:

```java
new Trigger(() -> Robot.getCurrentCoralTarget() == ReefTarget.L1)
    .and(new Trigger(not(arm::hasCoral)))
```

Just static import `not`.

spellingcat:
superseded by #11

PR #4, opened 2025-07-23
installing updated phoenix 

vivi-o:


src/main/java/frc/robot/Robot.java
```diff
@@ -8,7 +8,7 @@
 import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.CommandScheduler;
 
-public class Robot extends LoggedRobot {
+public class Robot extends TimedRobot {
```
spellingcat: so, there's a couple of ways to structure robot code. `TimedRobot` is like the most basic (usable) structure. a `LoggedRobot` is a `TimedRobot` + some other features that allow logging with advantagekit ([docs here](https://docs.advantagekit.org/getting-started/installation/existing-projects/#robot-configuration)), so when we want to use that additional logging functionality we need the robot to be a `LoggedRobot`

spellingcat (changes requested):
i believe main already has the line about loggedrobot, but yeah the phoenix vendordep needs to be updated

spellingcat (approved):
👍np

PR #6, opened 2025-07-29
Arm subsystem

vivi-o:


src/main/java/frc/robot/Robot.java
```diff
@@ -24,10 +27,13 @@ public enum RobotType {
...
       new ElevatorSubsystem(
           ROBOT_TYPE != RobotType.SIM ? new ElevatorIOReal() : new ElevatorIOSim());
-  private final ShoulderSubsystem shoulder = new ShoulderSubsystem();
-  private final IntakeSubsystem intake = new IntakeSubsystem();
 
-  private final Superstructure superstructure = new Superstructure(elevator, shoulder, intake);
+  private final ArmSubsystem arm =
+      new ArmSubsystem(ROBOT_TYPE != RobotType.SIM ? new ArmIOReal() : new ArmIOSim());
+  // private final ShoulderSubsystem shoulder = new ShoulderSubsystem();
+  // private final IntakeSubsystem intake = new IntakeSubsystem();
```
spellingcat: you can take out the shouldersubsystem stuff, since the arm subsystem takes care of the pivot and stuff

src/main/java/frc/robot/Superstructure.java
```diff
@@ -116,7 +117,7 @@ private Command forceState(SuperState nextState) {
 
   private void setSubstates() {
     elevator.setState(state.elevatorState);
-    shoulder.setState(state.shoulderState);
-    intake.setState(state.intakeState);
+    arm.setArmState(state.armState);
+    arm.setIntakeState(state.intakeState);
```
spellingcat: can i ask what the intake state does for the arm?

src/main/java/frc/robot/arm/ArmIO.java
```diff
@@ -0,0 +1,23 @@
+package frc.robot.arm;
+
+import edu.wpi.first.math.geometry.Rotation2d;
+import org.littletonrobotics.junction.AutoLog;
+
+public interface ArmIO {
+  @AutoLog
+  public static class ArmIOInputs {
+    public double angularVelocityRotsPerSec = 0.0;
+    public Rotation2d position = new Rotation2d();
```
spellingcat, line 10: the arm will also have a cancoder on it, so we'll want to log the position reported by the cancoder in addition to the position reported by the motor

spellingcat: `public Rotation2d cancoderPosition = new Rotation2d();`

src/main/java/frc/robot/arm/ArmIOReal.java
```diff
@@ -0,0 +1,96 @@
...
+  private final StatusSignal<AngularVelocity> angularVelocityRotsPerSec = motor.getVelocity();
+  private final StatusSignal<Current> supplyCurrentAmps = motor.getSupplyCurrent();
+  private final StatusSignal<Current> statorCurrentAmps = motor.getStatorCurrent();
+  private final StatusSignal<Voltage> appliedVoltage = motor.getMotorVoltage();
+  private final StatusSignal<Angle> motorPositionRotations = motor.getPosition();
+
+  private final VoltageOut voltageOut = new VoltageOut(0.0).withEnableFOC(true);
+  private final MotionMagicVoltage motionMagic = new MotionMagicVoltage(0.0).withEnableFOC(true);
+
+  public ArmIOReal() {
```
spellingcat, line 34: make sure to also add the cancoder

src/main/java/frc/robot/arm/ArmIOSim.java
```diff
@@ -0,0 +1,67 @@
...
+import edu.wpi.first.wpilibj.simulation.SingleJointedArmSim;
+
+public class ArmIOSim implements ArmIO {
+  // TODO: change to actual values
+  private final SingleJointedArmSim armSim =
+      new SingleJointedArmSim(
+          DCMotor.getKrakenX60Foc(1),
+          ArmSubsystem.PIVOT_RATIO, // pivot ratio
+          0.1,
+          Units.inchesToMeters(20), // arm length
```
spellingcat: we can worry about this later so this isn't blocking, but i believe the gear ratio is (44.0 / 16.0) * 23.0 and it's about 25 inches long

src/main/java/frc/robot/arm/ArmIOSim.java
```diff
@@ -0,0 +1,67 @@
...
+          ArmSubsystem.MIN_ANGLE.getRadians(), // min angle
+          ArmSubsystem.MAX_ANGLE.getRadians(), // max angle
+          true,
+          0.0);
+
+  private final ProfiledPIDController pivotPid =
+      new ProfiledPIDController(0.0, 0.0, 0.0, new TrapezoidProfile.Constraints(10.0, 10.0));
+  private final ArmFeedforward pivotFf = new ArmFeedforward(0.0, 0.0, 0.0);
+
```
spellingcat: i don't think this does anything because the gains are all 0

src/main/java/frc/robot/arm/ArmIOSim.java
```diff
@@ -0,0 +1,67 @@
...
+    appliedVoltage = voltage;
+    armSim.setInputVoltage(MathUtil.clamp(voltage, -20, 20));
+  }
+
+  @Override
+  public void setMotorPosition(Rotation2d targetPosition) {
+    setMotorVoltage(
+        pivotPid.calculate(armSim.getAngleRads(), targetPosition.getRadians())
+            + pivotFf.calculate(pivotPid.getSetpoint().position, pivotPid.getSetpoint().velocity));
+  }
```
spellingcat, line 58: same as above about the feedforward

src/main/java/frc/robot/arm/ArmIOSim.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @Override
+  public void setRollerVoltage(double voltage) {
+    // not gonna simulate the rollers
+  }
+
+  public void setResetSimState() {
+    armSim.setState(0.0, 0.0);
+  }
```
spellingcat: i don't think this gets called

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -0,0 +1,110 @@
...
+import frc.robot.Robot.RobotType;
+import java.util.function.Supplier;
+import org.littletonrobotics.junction.Logger;
+
+public class ArmSubsystem extends SubsystemBase {
+  // constants
+  // TODO: change to real values
+  public static final double PIVOT_RATIO = 10;
+  public static final Rotation2d MAX_ANGLE = Rotation2d.fromDegrees(180);
+  public static final Rotation2d MIN_ANGLE = Rotation2d.fromDegrees(0);
```
spellingcat: see previous comment about gear ratio, min/max angle is still tbd for now tho

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -0,0 +1,110 @@
...
+
+  private ArmIO io;
+  private ArmIOInputsAutoLogged inputs = new ArmIOInputsAutoLogged();
+
+  private ArmState armState = ArmState.IDLE;
+  private IntakeState intakeState = IntakeState.IDLE;
```
spellingcat: the intake state should be handled by the intake subsystem once that gets added, so there doesn't need to be any references to it in the arm stuff

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -0,0 +1,110 @@
...
+    // climbing
+    PRE_CLIMB(Rotation2d.fromDegrees(0)),
+    CLIMB(Rotation2d.fromDegrees(0));
+
+    public final Rotation2d position;
+
+    ArmState(Rotation2d position) {
+      this.position = position;
+    }
+  }
```
spellingcat: i would personally put this at the top of the file

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -0,0 +1,110 @@
...
+  // constants
+  // TODO: change to real values
+  public static final double PIVOT_RATIO = 10;
+  public static final Rotation2d MAX_ANGLE = Rotation2d.fromDegrees(180);
+  public static final Rotation2d MIN_ANGLE = Rotation2d.fromDegrees(0);
+
+  private ArmIO io;
+  private ArmIOInputsAutoLogged inputs = new ArmIOInputsAutoLogged();
+
+  private ArmState armState = ArmState.IDLE;
```
spellingcat:
```suggestion
  @AutoLogOutput(key = "Arm/State")
  private ArmState armState = ArmState.IDLE;
```

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -0,0 +1,110 @@
...
+  private IntakeState intakeState = IntakeState.IDLE;
+
+  public ArmSubsystem(ArmIO io) {
+    this.io = io;
+  }
+
+  @Override
+  public void periodic() {
+    io.updateInputs(inputs);
+    Logger.processInputs("arm", inputs);
```
spellingcat:
```suggestion
    Logger.processInputs("Arm", inputs);
```

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -0,0 +1,110 @@
...
+  public Command setTargetAngle(Rotation2d target) {
+    return setTargetAngle(() -> target);
+  }
+
+  public Command setTargetAngle(Supplier<Rotation2d> target) {
+    return this.runOnce(
+            () -> {
+              if (Robot.ROBOT_TYPE != RobotType.REAL) Logger.recordOutput("arm", target.get());
+            })
+        .andThen(this.run(() -> io.setMotorPosition(target.get())));
```
spellingcat:
```suggestion
    return this.runOnce(
            () -> {
              if (Robot.ROBOT_TYPE != RobotType.REAL) Logger.recordOutput("Arm", target.get());
              io.setMotorPosition(target.get())
            });
```
The `this.run()` would run the `setMotorPosition` method forever, which would then block everything else. so we can include it in the `runOnce()` with the logging statement. also make sure you have consistent capitalization

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -0,0 +1,110 @@
...
+    io.setRollerVoltage(voltage);
+  }
+
+  public void setArmState(ArmState state) {
+    this.armState = state;
+  }
+
+  public void setIntakeState(IntakeState state) {
+    this.intakeState = state;
+  }
```
spellingcat: see previous comment about intake

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -0,0 +1,110 @@
...
+    INTAKE(10.0), // prob add intake/outtake algae and intake coral as different states
+    OUTTAKE(-10.0);
+
+    public final double voltage;
+
+    IntakeState(double voltage) {
+      this.voltage = voltage;
+    }
+  }
+}
```
spellingcat: see previous comment about intake

src/main/java/frc/robot/Robot.java
```diff
@@ -24,10 +27,13 @@ public enum RobotType {
...
+
+  private final Superstructure superstructure = new Superstructure(elevator, arm);
 
   public Robot() {}
```
spellingcat, line 68: I can't comment on that line for some reason but also be sure to add `superstructure.periodic();` in `robotPeriodic()`, because the superstructure isn't a subsystem, the periodic method won't get called automatically, so we need to manually add it here

src/main/java/frc/robot/arm/ArmIO.java
```diff
@@ -0,0 +1,23 @@
...
+    public double statorCurrentAmps = 0.0;
+  }
+
+  void updateInputs(ArmIOInputs inputs);
+
+  void setMotorVoltage(double voltage);
+
+  void setMotorPosition(Rotation2d targetPosition);
+
+  void setRollerVoltage(double voltage);
```
spellingcat:
```suggestion
  public void updateInputs(ArmIOInputs inputs);

  public void setMotorVoltage(double voltage);

  public void setMotorPosition(Rotation2d targetPosition);

  public void setRollerVoltage(double voltage);
```
like, these are all *implicitly* public, but i'd feel a lot better if they were also marked as such

src/main/java/frc/robot/arm/ArmIOReal.java
```diff
@@ -0,0 +1,96 @@
...
+
+public class ArmIOReal implements ArmIO {
+  private final TalonFX motor = new TalonFX(14, "*");
+  private final TalonFX rollers = new TalonFX(15, "*");
+
+  private final StatusSignal<AngularVelocity> angularVelocityRotsPerSec = motor.getVelocity();
+  private final StatusSignal<Current> supplyCurrentAmps = motor.getSupplyCurrent();
+  private final StatusSignal<Current> statorCurrentAmps = motor.getStatorCurrent();
+  private final StatusSignal<Voltage> appliedVoltage = motor.getMotorVoltage();
+  private final StatusSignal<Angle> motorPositionRotations = motor.getPosition();
```
spellingcat: i think the motors + status signals should be declared here but instantiated inside the constructor

src/main/java/frc/robot/arm/ArmIOReal.java
```diff
@@ -0,0 +1,96 @@
...
+
+  private final VoltageOut voltageOut = new VoltageOut(0.0).withEnableFOC(true);
+  private final MotionMagicVoltage motionMagic = new MotionMagicVoltage(0.0).withEnableFOC(true);
+
```
spellingcat: i think that this should be `MotionMagicTorqueCurrentFOC` as per [this ctre article](https://v6.docs.ctr-electronics.com/en/latest/docs/api-reference/device-specific/talonfx/closed-loop-requests.html#choosing-output-type), i think basically it's just a bit more accurate

src/main/java/frc/robot/arm/ArmIOSim.java
```diff
@@ -0,0 +1,67 @@
...
+          ArmSubsystem.PIVOT_RATIO, // pivot ratio
+          0.1,
+          Units.inchesToMeters(20), // arm length
+          ArmSubsystem.MIN_ANGLE.getRadians(), // min angle
+          ArmSubsystem.MAX_ANGLE.getRadians(), // max angle
+          true,
+          0.0);
+
+  private final ProfiledPIDController pivotPid =
+      new ProfiledPIDController(0.0, 0.0, 0.0, new TrapezoidProfile.Constraints(10.0, 10.0));
```
spellingcat: also yeah the `ProfiledPIDController` numbers would need to be nonzero

spellingcat: lmk if you'd like help with tuning that (+ there should be a training article coming soon)

src/main/java/frc/robot/Robot.java
```diff
@@ -24,10 +27,13 @@ public enum RobotType {
...
```
spellingcat: make sure to add this inside `Robot() {}`

```
public Robot() {
  Logger.recordMetadata("ProjectName", "MyProject"); // Set a metadata value
  
  if (isReal()) {
      Logger.addDataReceiver(new WPILOGWriter()); // Log to a USB stick ("/U/logs")
      Logger.addDataReceiver(new NT4Publisher()); // Publish data to NetworkTables
  } else {
      setUseTiming(false); // Run as fast as possible
      String logPath = LogFileUtil.findReplayLog(); // Pull the replay log from AdvantageScope (or prompt the user)
      Logger.setReplaySource(new WPILOGReader(logPath)); // Read replay log
      Logger.addDataReceiver(new WPILOGWriter(LogFileUtil.addPathSuffix(logPath, "_sim"))); // Save outputs to a new log
  }
  
  Logger.start(); // Start logging! No more data receivers, replay sources, or metadata values may be added.
}
```

src/main/java/frc/robot/arm/ArmIOSim.java
```diff
@@ -0,0 +1,67 @@
...
+    inputs.position = Rotation2d.fromRadians(armSim.getAngleRads());
+    inputs.statorCurrentAmps = armSim.getCurrentDrawAmps();
+    inputs.supplyCurrentAmps = 0.0;
+    inputs.appliedVoltage = appliedVoltage;
+  }
+
+  @Override
+  public void setMotorVoltage(double voltage) {
+    appliedVoltage = voltage;
+    armSim.setInputVoltage(MathUtil.clamp(voltage, -20, 20));
```
spellingcat, line 50:
```suggestion
    armSim.setInputVoltage(voltage);
```
it'll clamp it for you, and our available voltage would be somewhere between 0 and 12

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  public void setMotorVoltage(double voltage) {
+    io.setMotorVoltage(voltage);
+  }
+
+  public void setMotorPosition(Rotation2d targetPosition) {
+    io.setMotorPosition(targetPosition);
+  }
+
```
spellingcat: i don't think these are needed for now because we're only ever using `setTargetAngle` to control the motor it seems

spellingcat (changes requested):
couple of changes (mostly about moving intake stuff out of the arm) but looking good so far, nice work :) let me know if you have any questions

src/main/java/frc/robot/Superstructure.java
vivi-o: The intake state was supposed to control the rollers on the end of the arm by setting the voltage, but now I have moved the logic into the intake subsystem.

src/main/java/frc/robot/Superstructure.java
```diff
@@ -26,26 +25,29 @@ public class Superstructure {
...
+    PRE_L3(ElevatorState.PRE_L3, ArmState.PRE_L3, IntakeState.IDLE),
+    L3(ElevatorState.L3, ArmState.L3, IntakeState.INTAKE_CORAL),
+    PRE_L4(ElevatorState.PRE_L4, ArmState.PRE_L4, IntakeState.IDLE),
+    L4(ElevatorState.L4, ArmState.L4, IntakeState.INTAKE_CORAL),
     POST_L4(ElevatorState.POST_L4, ArmState.POST_L4, IntakeState.IDLE),
     INTAKE_ALGAE_REEF_HIGH(
-        ElevatorState.INTAKE_ALGAE_REEF_HIGH, ArmState.INTAKE_ALGAE_REEF_HIGH, IntakeState.INTAKE),
+        ElevatorState.INTAKE_ALGAE_REEF_HIGH,
+        ArmState.INTAKE_ALGAE_REEF_HIGH,
+        IntakeState.INTAKE_ALGAE),
```
spellingcat: the intake won't be intaking algae, so during any of the intake algae stages it should just be in IDLE and look like [this](https://docs.google.com/document/d/1z6lj8qSADNNd7unU1ZCrEWAM7g3oJyfvDSwORe1Tlco/edit?tab=t.0#heading=h.h7zwam3gc0tw)

src/main/java/frc/robot/Superstructure.java
```diff
@@ -26,26 +25,29 @@ public class Superstructure {
...
     INTAKE_ALGAE_REEF_LOW(
-        ElevatorState.INTAKE_ALGAE_REEF_LOW, ArmState.INTAKE_ALGAE_REEF_LOW, IntakeState.INTAKE),
+        ElevatorState.INTAKE_ALGAE_REEF_LOW,
+        ArmState.INTAKE_ALGAE_REEF_LOW,
+        IntakeState.INTAKE_ALGAE),
     INTAKE_ALGAE_GROUND(
-        ElevatorState.INTAKE_ALGAE_GROUND, ArmState.INTAKE_ALGAE_GROUND, IntakeState.INTAKE),
-    BARGE(ElevatorState.BARGE, ArmState.BARGE, IntakeState.OUTTAKE),
+        ElevatorState.INTAKE_ALGAE_GROUND, ArmState.INTAKE_ALGAE_GROUND, IntakeState.INTAKE_ALGAE),
+    BARGE(ElevatorState.BARGE, ArmState.BARGE, IntakeState.OUTTAKE_ALGAE),
```
spellingcat: same with outtake

spellingcat: intake subsystem != rollers on the arm (those are controlled in the arm state)

src/main/java/frc/robot/arm/ArmIOReal.java
```diff
@@ -16,19 +17,30 @@
 import edu.wpi.first.units.measure.Voltage;
 
 public class ArmIOReal implements ArmIO {
-  private final TalonFX motor = new TalonFX(14, "*");
-  private final TalonFX rollers = new TalonFX(15, "*");
```
spellingcat: be sure to also have the rollers motor (+ all of its inputs and methods to control it)

src/main/java/frc/robot/arm/ArmIOReal.java
```diff
@@ -88,9 +93,4 @@ public void setMotorVoltage(double voltage) {
   public void setMotorPosition(Rotation2d targetPosition) {
     motor.setControl(motionMagic.withPosition(targetPosition.getRotations()));
   }
-
-  @Override
-  public void setRollerVoltage(double voltage) {
-    rollers.setControl(voltageOut.withOutput(voltage));
-  }
```
spellingcat: keep

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -6,63 +6,26 @@
...
-  }
-
-  public void setMotorPosition(Rotation2d targetPosition) {
-    io.setMotorPosition(targetPosition);
-  }
-
-  public void setRollerVoltage(double voltage) {
-    io.setRollerVoltage(voltage);
-  }
-
```
spellingcat: keep

spellingcat (changes requested):
agh sorry i should have been more clear. the arm *will* still have rollers on it, so there should be two motors--one to control the pivot, and one to control the wheels. the idea is that the `ArmState` enum stores a value for both of these (pivot angle and roller voltage) and we will need to control both of those in the Arm files

when i say intake, i'm referring to a wholly separate subsystem- the part that picks up (only) coral directly from the ground. this *also* has a pivot motor + rollers motor and the `IntakeState` will also capture both values, since we also need to control both of those, but this is handled in its own set of stuff. 
- there *is* a handoff between these two, but we'll worry about that later (and in any case that's a subsystem-subsystem interaction, not something that's wholly in one subsystem)
- another note is that the intake subsystem isn't involved with intaking algae at all, because that's just the arm picking the algae off the reef and it doesn't go through the intake or indexer

i'd focus on the arm for now and getting functionality implemented for both the rollers and the pivot. you're in a good place with it, just need to add a couple things back

really sorry for not being clearer, but hopefully this makes some sense 😭 you're doing great regardless tho! message me on slack if you've got any questions (even if they're "minor")!

vivi-o:
Hi Athena, 

Sorry, I got confused with the intake vs. arm rollers.  That's totally on me; I should've asked instead of guessing. Thanks for explaining it more clearly!

I also wanted to check in w/ you to ask if my current arm state implementation makes sense now (it includes both the arm position and roller voltage). I did some work on the intake subsystem, so should I focus on making that more accurate and functional next, or scrap it?

src/main/java/frc/robot/Robot.java
```diff
@@ -24,16 +31,38 @@ public enum RobotType {
...
+    } else {
+      setUseTiming(false); // Run as fast as possible
+      String logPath =
+          LogFileUtil
+              .findReplayLog(); // Pull the replay log from AdvantageScope (or prompt the user)
+      Logger.setReplaySource(new WPILOGReader(logPath)); // Read replay log
+      Logger.addDataReceiver(
+          new WPILOGWriter(
+              LogFileUtil.addPathSuffix(logPath, "_sim"))); // Save outputs to a new log
+    }
```
spellingcat:
```suggestion
switch (ROBOT_TYPE) {
      case REAL:
        Logger.addDataReceiver(new WPILOGWriter("/U")); // Log to a USB stick
        Logger.addDataReceiver(new NT4Publisher()); // Publish data to NetworkTables
        new PowerDistribution(1, ModuleType.kRev); // Enables power distribution logging
        break;
      case REPLAY:
        setUseTiming(false); // Run as fast as possible
        String logPath =
            LogFileUtil
                .findReplayLog(); // Pull the replay log from AdvantageScope (or prompt the user)
        Logger.setReplaySource(new WPILOGReader(logPath)); // Read replay log
        Logger.addDataReceiver(
            new WPILOGWriter(
                LogFileUtil.addPathSuffix(logPath, "_sim"))); // Save outputs to a new log
        break;
      case SIM:
        Logger.addDataReceiver(new NT4Publisher()); // Publish data to NetworkTables
        break;
    }
```

spellingcat (approved):
yeah no worries, I think this looks good for now! the arm state looks correct. as for the intake, i think it'll be easiest to pull a lot of the arm's roller+pivot functionality into a superclass and then have the intake inherit from that. 
that should go in a separate pr, and we can walk through that at tomorrow's meeting. if you're around between like 3:30-4 i can probably get you started then before i have to talk to the new people but if not i can also just talk to you while they're working. nice work :D

spellingcat:
intake is breaking build rn but it's ok

PR #9, opened 2025-09-16
Swerve

SCool62:
Here's my swerve rewrite. Teleop driving is tested in sim, but not in real life. I haven't tested closed-loop control yet. LMK if there's any parts that are still hard to read/understand.


src/main/java/frc/robot/Robot.java
```diff
@@ -4,23 +4,67 @@
...
+
+    private RobotHardware(SwerveConstants constants) {
+      this.constants = constants;
+    }
+
+    public SwerveConstants getSwerveConstants() {
+      return constants;
+    }
+  }
+
```
spellingcat: probably not gonna do this for this robot since we're only going to be running it on one bot and i'm not sure if we'll have an alpha bot in season

src/main/java/frc/robot/Robot.java
```diff
@@ -29,13 +73,124 @@ public enum RobotType {
...
+                  ROBOT_HARDWARE.getSwerveConstants().getDriveGearRatio(),
+                  ROBOT_HARDWARE.getSwerveConstants().getTurnGearRatio(),
+                  // These friction voltages are copied from Reefscape repo
+                  Volts.of(0.1),
+                  Volts.of(0.2),
+                  Meter.of(ROBOT_HARDWARE.getSwerveConstants().getWheelRadiusMeters()),
+                  // Copied from Reefscape
+                  KilogramSquareMeters.of(0.03),
+                  // Copied from Reefscape
+                  1.5))
```
spellingcat: do you know how you'd find these numbers?

spellingcat: not sure if it's something we can calculate from cad but worth asking them

src/main/java/frc/robot/Robot.java
```diff
@@ -29,13 +73,124 @@ public enum RobotType {
 
   private final Superstructure superstructure = new Superstructure(elevator, shoulder, intake);
 
-  public Robot() {}
+  // Maple Sim Stuff
+  private final DriveTrainSimulationConfig driveTrainSimConfig =
```
spellingcat, line 232: may need to be optional?

src/main/java/frc/robot/Robot.java
```diff
@@ -29,13 +73,124 @@ public enum RobotType {
...
+              new SwerveDriveSimulation(driveTrainSimConfig, new Pose2d(3, 3, Rotation2d.kZero)))
+          : Optional.empty();
+
+  // Subsystem initialization
+  private final SwerveSubsystem swerve =
+      new SwerveSubsystem(
+          ROBOT_HARDWARE.getSwerveConstants(),
+          ROBOT_TYPE != RobotType.SIM
+              ? new GyroIOReal(ROBOT_HARDWARE.getSwerveConstants().getGyroID())
+              : new GyroIOSim(swerveSimulation.get().getGyroSimulation()),
```
spellingcat: could probably move the gyro stuff inside swerve too, same as the modules

src/main/java/frc/robot/Robot.java
```diff
@@ -29,13 +73,124 @@ public enum RobotType {
...
+        break;
+    }
+
+    Logger.start(); // Start logging! No more data receivers, replay sources, or metadata values may
+    // be added.
+
+    if (ROBOT_TYPE == RobotType.SIM) {
+      SimulatedArena.getInstance().addDriveTrainSimulation(swerveSimulation.get());
+    }
+
```
spellingcat, line 333: you'll need to add `swerve.startOdoThread();`

src/main/java/frc/robot/Robot.java
```diff
@@ -29,13 +73,124 @@ public enum RobotType {
...
+      case SIM:
+        Logger.addDataReceiver(new NT4Publisher()); // Publish data to NetworkTables
+        break;
+    }
+
+    Logger.start(); // Start logging! No more data receivers, replay sources, or metadata values may
+    // be added.
+
+    if (ROBOT_TYPE == RobotType.SIM) {
+      SimulatedArena.getInstance().addDriveTrainSimulation(swerveSimulation.get());
```
spellingcat: possible no such element exception

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -0,0 +1,392 @@
...
+    this.odometryThread = odometryThread;
+  }
+
+  public void startOdoThread() {
+    odometryThread.start();
+  }
+
+  @Override
+  public void periodic() {
+    Tracer.trace(
```
spellingcat, line 167: tracer stuff lowkey might need to be refactored/trace more things later, not an action item for you but just a Thing

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -0,0 +1,392 @@
...
+          Tracer.trace("Update gyro inputs", () -> gyroIO.updateInputs(gyroInputs));
+          Logger.processInputs("Swerve/Gyro", gyroInputs);
+
+          for (Module module : modules) {
+            Tracer.trace("Update module inputs for " + module.getPrefix(), module::periodic);
+          }
+
+          Tracer.trace("Update odometry", this::updateOdometry);
+        });
+  }
```
spellingcat, line 197: i think you need something to stop the modules on disable

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -0,0 +1,392 @@
...
+
+  private void drive(ChassisSpeeds speeds, boolean openLoop) {
+    // Converts time continuous chassis speeds to setpoints after the specified time (dtSeconds)
```
spellingcat: add javadoc

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -0,0 +1,392 @@
...
+    SwerveDriveKinematics.desaturateWheelSpeeds(states, swerveConstants.getMaxLinearSpeed());
+    if (Robot.ROBOT_TYPE != RobotType.REAL) Logger.recordOutput("SwerveStates/Setpoints", states);
+
+    if (Robot.ROBOT_TYPE != RobotType.REAL) Logger.recordOutput("Swerve/Target Speeds", speeds);
+
+    SwerveModuleState[] optimizedStates = new SwerveModuleState[modules.length];
+
+    for (int i = 0; i < optimizedStates.length; i++) {
+      if (openLoop) {
+        optimizedStates[i] = modules[i].runOpenLoop(states[i], true);
```
spellingcat: can you explain why you dropped the `focEnable` check?

src/main/java/frc/robot/swerve/module/Module.java
```diff
@@ -0,0 +1,113 @@
...
+   * Runs closed-loop to the specified state. Used for automated actions, i.e. auto, autoalign etc.
+   *
+   * @param state setpoint state
+   * @return the optimized state
+   */
+  public SwerveModuleState runClosedLoop(SwerveModuleState state) {
+    state.optimize(getAngle());
+
+    io.setTurnSetpoint(state.angle);
+    io.setDriveSetpoint(state.speedMetersPerSecond);
```
spellingcat: i would perhaps rename this to `setDriveVelocitySetpoint`

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -0,0 +1,392 @@
...
+
+        lastModulePositions[moduleIndex] = modulePositions[moduleIndex];
+      }
+
+      if (hasNullModulePosition) {
+        missingModuleData.set(true);
+        if (!gyroInputs.isConnected
+            || sample
+                    .values()
+                    .get(new SignalID(SignalType.GYRO, PhoenixOdometryThread.GYRO_MODULE_ID))
```
spellingcat: i wonder if making all these new `SignalID`s leads to any memory/gc issues?

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -0,0 +1,392 @@
...
+
+    if (Robot.ROBOT_TYPE != RobotType.REAL)
+      Logger.recordOutput("SwerveStates/SetpointsOptimized", optimizedStates);
+  }
+
+  private void updateOdometry() {
+
+    List<Samples> sampledStates = odometryThreadInputs.sampledStates;
+    // Use sync samples if there aren't any async ones
+    if (sampledStates.size() == 0) {
```
spellingcat: you also need to check if the robot is sim, or if the list isn't empty but the samples are iirc

src/main/java/frc/robot/swerve/SwerveSubsystem.java
spellingcat: maybe you could move this into its own check - analogous to the `nullModulePositions` - and then check each combo of missing gyro/missing module data

spellingcat: perhaps not strictly necessary but might be easier to read (and debug later if needed)

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -0,0 +1,392 @@
...
+
+  /** Returns the pose estimator rotation, as returned by {@link #getPose()} */
+  public Rotation2d getRotation() {
+    return getPose().getRotation();
+  }
+
+  public Rotation3d getRotation3d() {
+    return new Rotation3d(
+        gyroInputs.roll.getRadians(), gyroInputs.pitch.getRadians(), gyroInputs.yaw.getRadians());
+  }
```
spellingcat: you can delete this

src/main/java/frc/robot/Robot.java
spellingcat: nvm

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -0,0 +1,392 @@
...
+  /**
+   * Drive at a robot-relative speed open-loop.
+   *
+   * @param speeds the robot-relative speed setpoint.
+   * @return a Command driving to the target speeds.
+   */
+  public Command driveOpenLoop(Supplier<ChassisSpeeds> speeds) {
+    return this.run(() -> drive(speeds.get(), true));
+  }
+
```
spellingcat: this isn't used

spellingcat: i suggest renaming `driveTeleop` to `driveOpenLoop` and noting in the javadoc that this is what we use for teleop

spellingcat: or otherwise standardizing between teleop/open loop and auto/closed loop

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -0,0 +1,392 @@
...
+   * Drive closed-loop at a field relative speed
+   *
+   * @param speeds the field-relative speed setpoint
+   * @return a Command driving to those speeds
+   */
+  public Command driveVelocityFieldRelative(Supplier<ChassisSpeeds> speeds) {
+    return driveClosedLoop(
+        () -> ChassisSpeeds.fromFieldRelativeSpeeds(speeds.get(), getRotation()));
+  }
+
```
spellingcat: don't think we ever drive field relative

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -0,0 +1,392 @@
...
+          for (Module module : modules) {
+            Tracer.trace("Update module inputs for " + module.getPrefix(), module::periodic);
+          }
+
+          Tracer.trace("Update odometry", this::updateOdometry);
+        });
+  }
+
+  private void drive(ChassisSpeeds speeds, boolean openLoop) {
```
spellingcat: really nitpicky but i might move the drive command to be after the update odo stuff just to keep all the drive related stuff together

spellingcat: and also move the getters to the end

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -0,0 +1,392 @@
...
+
+  /**
+   * Drive closed-loop at robot relative speeds
+   *
+   * @param speeds robot relative speed setpoint
+   * @return a command driving to target speeds
+   */
+  public Command driveClosedLoop(Supplier<ChassisSpeeds> speeds) {
+    return this.run(() -> drive(getVelocityFieldRelative(), false));
+  }
```
spellingcat: change `getVelocityFieldRelative()` to the `speeds`parameter

src/main/java/frc/robot/swerve/constants/KelpieSwerveConstants.java
```diff
@@ -0,0 +1,162 @@
...
+
+  @Override
+  public ModuleConstants getBackLeftModule() {
+    return new ModuleConstants(2, "Back Left", 4, 5, 2, Rotation2d.fromRotations(-0.276855));
+  }
+
+  @Override
+  public ModuleConstants getBackRightModule() {
+    return new ModuleConstants(3, "Back Right", 6, 7, 3, Rotation2d.fromRotations(-0.211426));
+  }
```
spellingcat: nitpick but can we move these to be with the other module getters

src/main/java/frc/robot/swerve/constants/KelpieSwerveConstants.java
```diff
@@ -0,0 +1,162 @@
...
+  @Override
+  public double getDriveGearRatio() {
+    // Taken from https://www.swervedrivespecialties.com/products/mk4n-swerve-module, L2+
+    // configuration
+    return (50.0 / 16.0) * (17.0 / 27.0) * (45.0 / 15.0);
+  }
+
+  @Override
+  public double getTurnGearRatio() {
+    // For SDS Mk4n
```
spellingcat, line 183: fyi that the offseason bot is gonna be mk4i but we need to redo much of this anyway

src/main/java/frc/robot/swerve/constants/SwerveConstants.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  // Hardware constants
+  public abstract ModuleConstants getFrontLeftModule();
+
+  public abstract ModuleConstants getFrontRightModule();
+
+  public abstract ModuleConstants getBackLeftModule();
+
+  public abstract ModuleConstants getBackRightModule();
```
spellingcat: maybe rename to `getXModuleConstants`

src/main/java/frc/robot/swerve/gyro/GyroIOReal.java
```diff
@@ -0,0 +1,60 @@
+package frc.robot.swerve.gyro;
```
spellingcat, line 1: ty for putting these in folders btw

src/main/java/frc/robot/swerve/gyro/GyroIOReal.java
```diff
@@ -0,0 +1,60 @@
...
+
+    yaw = pigeon.getYaw();
+    pitch = pigeon.getPitch();
+    roll = pigeon.getRoll();
+    yawVelocity = pigeon.getAngularVelocityZWorld();
+
+    BaseStatusSignal.setUpdateFrequencyForAll(50.0, pitch, roll);
+    BaseStatusSignal.setUpdateFrequencyForAll(
+        PhoenixOdometryThread.ODOMETRY_FREQUENCY_HZ, yaw, yawVelocity);
+    // TODO: ASK IF yawVelocity SHOULD BE REGISTERED TOO
```
spellingcat: hmm ig we could although i don't really think it's used for much

spellingcat: also means you probably don't need it to be at the higher frequency

src/main/java/frc/robot/swerve/module/Module.java
spellingcat: to be clear i'm talking about line 39 (so rename it in io)

src/main/java/frc/robot/swerve/module/ModuleIO.java
```diff
@@ -0,0 +1,45 @@
...
+
+  public void updateInputs(ModuleIOInputs inputs);
+
+  public void setDriveVoltage(double volts, boolean withFoc);
+
+  public default void setDriveVoltage(double volts) {
+    setDriveVoltage(volts, true);
+  }
+
+  public void setDriveSetpoint(double setpointMetersPerSecond);
```
spellingcat: `setDriveVelocitySetpoint`

src/main/java/frc/robot/swerve/module/Module.java
```diff
@@ -0,0 +1,113 @@
...
+  public SwerveModulePosition getPosition() {
+    return new SwerveModulePosition(getPositionMeters(), getAngle());
+  }
+
+  /** Returns the module state (turn angle and drive velocity) at normal sampling frequency. */
+  public SwerveModuleState getState() {
+    return new SwerveModuleState(getVelocityMetersPerSec(), getAngle());
+  }
+
+  public void setTurnSetpoint(Rotation2d rotation) {
```
spellingcat, line 112: `setTurnPositionSetpoint`

spellingcat: err change it in io too (i'm losing track of where i've put comments)

src/main/java/frc/robot/swerve/module/Module.java
```diff
@@ -0,0 +1,113 @@
...
+  private void runVoltageSetpoint(double volts, Rotation2d targetAngle, boolean focEnabled) {
+    io.setTurnSetpoint(targetAngle);
+    io.setDriveVoltage(
+        // I think this is supposed avoid moving the drive too much when the turn motor is out of
+        // position
+        volts * Math.cos(targetAngle.minus(inputs.turnPosition).getRadians()), focEnabled);
+  }
+
+  /** Returns the current turn angle of the module at normal sampling frequency. */
+  public Rotation2d getAngle() {
```
spellingcat, line 80: forgot what the deal is with this normal sampling frequency thing - can you check if we ever actually need to make that distinction or not? you could look at reefscape if needed

spellingcat: tbh i think i was the one who wrote that so i can also look at it instead if you need

src/main/java/frc/robot/swerve/module/ModuleIOReal.java
```diff
@@ -0,0 +1,186 @@
...
+        PhoenixOdometryThread.ODOMETRY_FREQUENCY_HZ, drivePosition, turnPosition);
+
+    driveTalon.optimizeBusUtilization();
+    turnTalon.optimizeBusUtilization();
+    cancoder.optimizeBusUtilization();
+  }
+
+  @Override
+  public void updateInputs(ModuleIOInputs inputs) {
+    BaseStatusSignal.refreshAll(
```
spellingcat: i'd still log this statuscode

src/main/java/frc/robot/swerve/module/ModuleIOReal.java
```diff
@@ -0,0 +1,186 @@
...
+    inputs.turnPosition = Rotation2d.fromRotations(turnPosition.getValueAsDouble());
+    inputs.turnVelocityRadPerSec = turnVelocity.getValue().in(RadiansPerSecond);
+    inputs.turnAppliedVolts = turnAppliedVolts.getValueAsDouble();
+    inputs.turnTempC = turnTemp.getValueAsDouble();
+    inputs.turnStatorCurrentAmps = turnStatorCurrent.getValueAsDouble();
+    inputs.turnSupplyCurrentAmps = turnSupplyCurrent.getValueAsDouble();
+  }
+
+  @Override
+  public void setDriveVoltage(double volts, boolean withFoc) {
```
spellingcat, line 166: kinda random and idk if it's actually a problem but i'd double check if your usage of the foc enable thing is consistent or has stuff hardcoded. if it isn't already, i'm thinking it should all eventually mapped back to one variable to make sure it's consistent if it ever does change

src/main/java/frc/robot/swerve/module/Module.java
```diff
@@ -0,0 +1,113 @@
...
+    runVoltageSetpoint(volts, state.angle, focEnabled);
+
+    return state;
+  }
+
+  private void runVoltageSetpoint(double volts, Rotation2d targetAngle, boolean focEnabled) {
+    io.setTurnSetpoint(targetAngle);
+    io.setDriveVoltage(
+        // I think this is supposed avoid moving the drive too much when the turn motor is out of
+        // position
```
spellingcat, line 75: i _think_ what this does is scale the drive voltage output by how "off" it is. for example if `inputs.turnPosition` matches `targetAngle` perfectly, `targetAngle.minus(inputs.turnPosition)` is going to equal 0 and cos(0) = 1, which means you're putting volts*1 or the full voltage out. versus if it's 90 degrees off, cos(90) = 0 and so there's no drive voltage since it has to turn

spellingcat: as to why that's necessary i'm not exactly sure of the math, but i think it has something to do with avoiding drifting when rotating and translating simultaneously

src/main/java/frc/robot/swerve/odometry/OdometryThreadIO.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public void toLog(LogTable table) {
+      final var timestamps = sampledStates.stream().mapToDouble(Samples::timestamp).toArray();
+      final Set<Integer> modIds = new HashSet();
+      for (int i = 0; i < sampledStates.size(); i++) {
+        final var sample = sampledStates.get(i);
+        for (var signal : sample.values().entrySet()) {
+          table.put(
+              "Data/" + i + " " + signal.getKey().type().toString() + " " + signal.getKey().modID(),
```
spellingcat, line 28: maybe add "module" before modID (or the actual prefix- maybe jsut swap them all for the prefix?)

src/main/java/frc/robot/swerve/odometry/OdometryThreadIO.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public void fromLog(LogTable table) {
+      sampledStates = List.of();
+      final var modIds = table.get("Module IDs").getIntegerArray();
+      final var timestamps = table.get("Timestamps").getDoubleArray();
+      for (int i = 0; i < sampledStates.size(); i++) {
+        var timestamp = timestamps[i];
+        try {
+          Map<SignalID, Double> values = new HashMap();
+          for (var id : modIds) {
```
spellingcat, line 48: i lied don't get rid of the modids

spellingcat: still, maybe there's a way to make this more readable with the prefixes. not blocking tho

src/main/java/frc/robot/swerve/odometry/PhoenixOdometryThread.java
```diff
@@ -0,0 +1,212 @@
...
+
+  /** Represents a new signal to be registered on the thread */
+  public record Registration(
+      ParentDevice device,
+      Optional<ModuleConstants> moduleConstants,
+      SignalType type,
+      Set<BaseStatusSignal> signals) {}
+
+  /** modID should be GYRO_MODULE_ID for the gyro signal */
+  public record RegisteredSignal(BaseStatusSignal signal, int modID, SignalType type) {}
```
spellingcat: seems odd that the registered signal doesn't use a signal id for the type and modid

src/main/java/frc/robot/swerve/odometry/PhoenixOdometryThread.java
```diff
@@ -0,0 +1,212 @@
...
+    GYRO;
+    // In theory we could measure other types of info in this thread
+    // But we don't!
+  }
+
+  /** modID should be OdometryThreadIO.GYRO_MODULE_ID for the gyro signal */
+  public record SignalID(SignalType type, int modID) {}
+
+  /** Represents a new signal to be registered on the thread */
+  public record Registration(
```
spellingcat, line 38: maybe rename to signal (though maybe that runs into issues with the ctre signals?) idk, i think the name Registration is a little odd

src/main/java/frc/robot/swerve/odometry/PhoenixOdometryThread.java
```diff
@@ -0,0 +1,212 @@
...
+                                ? registration.moduleConstants().get().id()
+                                : GYRO_MODULE_ID,
+                            registration.type()))
+                .toList());
+        registration.signals.stream()
+            .forEach(
+                (signal) -> {
+                  // Grow the array and copy the signal into the new array slot
+                  signalArr = Arrays.copyOf(signalArr, signalArr.length + 1);
+                  signalArr[signalArr.length - 1] = signal;
```
spellingcat, line 111: i wonder if this really needs to be an array and not an arraylist?

spellingcat: like, the only place i see an array being necessary is on L161 and there you could just use `.toArray()`

spellingcat: totally possible i'm missing something tho

src/main/java/frc/robot/swerve/odometry/PhoenixOdometryThread.java
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  private PhoenixOdometryThread() {
+    // EvictingQueue removes the oldest entry after the max size has been reached
+    this(EvictingQueue.create(5));
+  }
+
+  public void registerSignals(Collection<Registration> registrations) {
+    registerSignals(registrations.toArray(new Registration[] {}));
+  }
```
spellingcat: not used

src/main/java/frc/robot/swerve/odometry/PhoenixOdometryThread.java
```diff
@@ -0,0 +1,212 @@
...
+        assert new CANBus(registration.device.getNetwork()).isNetworkFD()
+            : "Only CAN FDs supported";
+        // Add each Registration to the Registered Signals array so it can be updated in the thread
+        registeredSignals.addAll(
+            registration.signals.stream()
+                .map(
+                    signal ->
+                        new RegisteredSignal(
+                            signal,
+                            // If there's no module constants, we must be reading from the Gyro
```
spellingcat: good

src/main/java/frc/robot/swerve/odometry/PhoenixOdometryThread.java
```diff
@@ -0,0 +1,212 @@
...
+                            registration.moduleConstants().isPresent()
+                                ? registration.moduleConstants().get().id()
+                                : GYRO_MODULE_ID,
+                            registration.type()))
+                .toList());
+        registration.signals.stream()
+            .forEach(
+                (signal) -> {
```
spellingcat, line 108: be the change

src/main/java/frc/robot/swerve/odometry/PhoenixOdometryThread.java
```diff
@@ -0,0 +1,212 @@
...
+        () -> {
+          var readLock = journalLock.readLock();
+          try {
+            readLock.lock();
+
+            return Tracer.trace(
+                "stream timestamps",
+                () ->
+                    journal.stream()
+                        .filter(s -> s.timestamp > timestamp)
```
spellingcat: .filter(signal -> signal.timestamp > timestamp)

src/main/java/frc/robot/util/MaplePhoenixUtil.java
```diff
@@ -0,0 +1,92 @@
+package frc.robot.util;
```
spellingcat: not sure if you copied this from reefscape but it appears that we copied it from 6328 and they have a license at the top so prob want to keep that in, but i'm not really worried about it

spellingcat (changes requested):
very, very nice work overall. some relatively minor things

src/main/java/frc/robot/Robot.java
SCool62: K. I'm removing the enum, but i still think its useful to have the constants in 1 class so i'm putting them as a constant in the swerve subsystem

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: Reefscape swerve had this in periodic, but i've made it a command that triggers when the robot is disabled. Seems to me like the more "correct" way to do it but lmk if you think its a problem.

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: oops mb

SCool62: oops mb

src/main/java/frc/robot/swerve/module/Module.java
SCool62: yeah that makes sense

src/main/java/frc/robot/swerve/constants/KelpieSwerveConstants.java
SCool62: yeah. we'll need to redo basically the whole swerve constants file

src/main/java/frc/robot/swerve/gyro/GyroIOReal.java
SCool62: No problem

src/main/java/frc/robot/swerve/module/Module.java
SCool62: could be wrong but i think the distinction is that getAngle() returns the position being sampled with sync odometry, so might lag behind the async odometry thread

src/main/java/frc/robot/swerve/module/Module.java
SCool62: yeah that was my understanding too. it probably helps it reach the setpoint more consistently (because the drive motor moves less when the wheel isn't oriented)

src/main/java/frc/robot/swerve/odometry/PhoenixOdometryThread.java
SCool62: I think registration makes sense bc it represents all the data required to register a signal but i can see how it could be confusing

src/main/java/frc/robot/swerve/odometry/PhoenixOdometryThread.java
SCool62: Could do that although each time you call toArray(), it either copies the array or uses a generator to make a new one. might end up making a lot of objects in the loop

src/main/java/frc/robot/swerve/gyro/GyroIOReal.java
SCool62: yeah we don't really use it i think. i'm going to update it with the rest of 50Hz odo unless you think otherwise

src/main/java/frc/robot/Robot.java
SCool62: its not in our cad. i'm looking at the SDS cad but idk if it'll be there

SCool62: honestly not sure why i don't just use a COTS swerve sim (they have one for mk4 and variations, not mk5 yet tho)

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: Switched this out for a constant.

src/main/java/frc/robot/Robot.java
SCool62: I can make it optional but I don't see why it has to be. Its just a constant configuration so I see no reason to make it optional.

src/main/java/frc/robot/swerve/module/ModuleIOReal.java
SCool62: wdym. I think the foc is controlled by a heuristic in drive(), which I forgot to add back initally.

src/main/java/frc/robot/swerve/odometry/PhoenixOdometryThread.java
spellingcat: isn't this basically what's happening here tho? with the copy of

src/main/java/frc/robot/swerve/module/ModuleIOReal.java
spellingcat: like i'd recommend you just go through and make sure you're not accidentally hardcoding any "true"s

src/main/java/frc/robot/swerve/module/Module.java
spellingcat: yep, could you check if/where we ever actually use the sync one tho

src/main/java/frc/robot/Robot.java
```diff
@@ -29,13 +57,114 @@ public enum RobotType {
...
+                new ChassisSpeeds(
+                        modifyJoystick(driver.getLeftY())
+                            * SwerveSubsystem.SWERVE_CONSTANTS.getMaxLinearSpeed(),
+                        modifyJoystick(driver.getLeftX())
+                            * SwerveSubsystem.SWERVE_CONSTANTS.getMaxLinearSpeed(),
+                        modifyJoystick(driver.getRightX())
+                            * SwerveSubsystem.SWERVE_CONSTANTS.getMaxAngularSpeed())
+                    .times(-1)));
+    
+    new Trigger(this::isDisabled).whileTrue(swerve.stop());
```
spellingcat:
```suggestion
    RobotModeTriggers.disabled().whileTrue(swerve.stop());
```

src/main/java/frc/robot/Robot.java
spellingcat: does maplesim have that?

src/main/java/frc/robot/Robot.java
spellingcat: yeah mb got confused with the swerve sim

spellingcat: (does that one need to be optional either? i feel like we're always checking if it's sim around it)

src/main/java/frc/robot/Robot.java
SCool62: yeah its built in

SCool62: yeah its built in

src/main/java/frc/robot/Robot.java
SCool62: i guess it doesn't. it looks like it just assigns stuff in the constructor. i'll make it not optional

src/main/java/frc/robot/swerve/module/Module.java
SCool62: we use getAngle() primarily to optimize the swerve setpoints. i don't think it should be removed.

src/main/java/frc/robot/swerve/module/ModuleIOReal.java
SCool62: Looks like the only place where we use the setDriveVoltage() method is in SwerveSubsystem.drive(), with the heuristic.

src/main/java/frc/robot/swerve/odometry/PhoenixOdometryThread.java
SCool62: No. This only makes a copy when you register a new signal (only happens like once for the gyro and each module). If I used an ArrayList, and use toArray() in the waitForUpdate(), it would make a copy of the array every loop cycle I think.

src/main/java/frc/robot/swerve/constants/KelpieSwerveConstants.java
```diff
@@ -116,6 +119,12 @@ public double getMaxLinearAcceleration() {
     return 14.0;
   }
 
+  @Override
+  public Mass getMass() {
+    // With old end effector (new EE isn't in the main robot cad...)
```
spellingcat, line 139: ?

src/main/java/frc/robot/swerve/module/ModuleIOReal.java
spellingcat: kk cool. i know there's a couple of other places that use a boolean for foc or not, so mostly just wanted to check that it's *always* the same everywhere

src/main/java/frc/robot/swerve/odometry/PhoenixOdometryThread.java
spellingcat: ah, makes sense

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -221,57 +213,123 @@ private void updateOdometry() {
 
         lastModulePositions[moduleIndex] = modulePositions[moduleIndex];
       }
+      if (!gyroInputs.isConnected || sample.values().get(GYRO_SIGNAL_ID) == null) {
+        hasNullGyroRotation = true;
+      } else {
+        hasNullGyroRotation = false;
+      }
```
spellingcat: could make this a ternary

src/main/java/frc/robot/Robot.java
```diff
@@ -79,41 +58,25 @@ public SwerveConstants getSwerveConstants() {
...
-                  // These friction voltages are copied from Reefscape repo
-                  Volts.of(0.1),
-                  Volts.of(0.2),
-                  Meter.of(ROBOT_HARDWARE.getSwerveConstants().getWheelRadiusMeters()),
-                  // Copied from Reefscape
-                  KilogramSquareMeters.of(0.03),
-                  // Copied from Reefscape
-                  1.5))
+                  // Still not sure where the 1.5 came from
+                  1.5,
```
spellingcat, line 241: it's the coefficient of friction (which *can* be greater than 1 for really grippy stuff (like wheels on a carpet!) that's "harder to push than lift". i remember people got in an argument about this last year lol)

spellingcat: as for how we measured that i have no clue

spellingcat (commented):
lgtm. once this has been verified irl i'll approve

spellingcat:
oh also merge main in when you get a chance. superstructure got merged and also has some assorted autoaim methods that it would be great if you could fill out 🙏

src/main/java/frc/robot/swerve/constants/KelpieSwerveConstants.java
SCool62: Old EE has slightly lower mass but idt it matters that much. I suppose if nessesary I could remove old EE from mass calculation and add back the mass of the new EE

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: or could just assign the boolean directly.

src/main/java/frc/robot/Robot.java
```diff
@@ -92,7 +107,32 @@ public static enum ScoringSide {
...
+              Meter.of(SwerveSubsystem.SWERVE_CONSTANTS.getTrackWidthY()))
+          .withBumperSize(
+              Meter.of(SwerveSubsystem.SWERVE_CONSTANTS.getBumperWidth()),
+              Meter.of(SwerveSubsystem.SWERVE_CONSTANTS.getBumperLength()))
+          .withRobotMass(SwerveSubsystem.SWERVE_CONSTANTS.getMass());
+
+  private final SwerveDriveSimulation swerveSimulation =
+      new SwerveDriveSimulation(driveTrainSimConfig, new Pose2d(3, 3, Rotation2d.kZero));
+  // Subsystem initialization
+  private final SwerveSubsystem swerve = new SwerveSubsystem(swerveSimulation, this::getSuperstructureState);
```
spellingcat, line 256: this might not work because you're calling it before superstructure is defined? it's kind of unpleasant but maybe just make a new `setStateSupplier` method that sets the variable there (and not in the constructor) in swerve and then call that in the `Robot()` constructor

src/main/java/frc/robot/camera/Camera.java
```diff
@@ -0,0 +1,262 @@
...
+    this.io.setSimPose(simEst, newResult);
+  }
+
+  public String getName() {
+    return io.getName();
+  }
+
+  public static Matrix<N3, N1> findVisionMeasurementStdDevs(EstimatedRobotPose estimation) {
+    double sumDistance = 0;
+    for (var target : estimation.targetsUsed) {
```
spellingcat: ik you are quite passionate about this issue so if you would be so kind as to change the `var`s that would be wonderful

src/main/java/frc/robot/camera/Camera.java
```diff
@@ -0,0 +1,262 @@
...
+                                      .anyMatch(
+                                          t ->
+                                              t.getFiducialId() == 4
+                                                  || t.getFiducialId() == 5
+                                                  || t.getFiducialId() == 15
+                                                  || t.getFiducialId() == 14)
+                                  ? 1.2
+                                  : 1.0)
+                          .times(stateSupplier.get() == SuperState.PRE_BARGE_LEFT || stateSupplier.get() == SuperState.PRE_BARGE_RIGHT ? 0.5 : 1.0));
+              // the sussifier
```
spellingcat:
```suggestion
```

spellingcat: camera names need to change but that's a bring up problem! put a `TODO` there for now

src/main/java/frc/robot/swerve/constants/SwerveConstants.java
```diff
@@ -0,0 +1,146 @@
...
+public abstract class SwerveConstants {
+
+  protected static final Alert multipleInstancesAlert =
+      new Alert("Multiple Swerve Constants Files", AlertType.kError);
+  private static final Alert tagLoadFailureAlert =
+      new Alert("Failed to load custom tag map", AlertType.kWarning);
+  protected AprilTagFieldLayout fieldTags;
+
+  public SwerveConstants() {
+    try {
```
spellingcat, line 27: for some context in case you were wondering- this is here because at one point in time we wanted to run some sort of tagslam and modify the field layout in case stuff moved throughout the match. that didn't happen so this is not strictly necessary (but hey if you're hurting for things to do that's a thing!)

src/main/java/frc/robot/Robot.java
```diff
@@ -336,7 +403,21 @@ private void addAutos() {
   @Override
   public void robotPeriodic() {
     CommandScheduler.getInstance().run();
-    superstructure.periodic();
+    // superstructure.periodic();
```
spellingcat: keep

spellingcat (changes requested):
almost there yippee

src/main/java/frc/robot/camera/Camera.java
SCool62: TODO is in Offseason bot constants

src/main/java/frc/robot/Robot.java
SCool62: I could do that, but it shouldn't be a problem bc the supplier won't get called until the first Command scheduler run which happens after both are instantiated. I'm not a huge fan either tho. Other option would be to make state static.

SCool62: I'm going to make it static

src/main/java/frc/robot/camera/Camera.java
```diff
@@ -179,12 +178,12 @@ public void updateCamera(SwerveDrivePoseEstimator swerveEstimator) {
                           .times(
                               (getName().equals("Front_Left_Camera")
                                           || getName().equals("Front_Right_Camera"))
-                                      && (stateSupplier.get().toString().startsWith("PRE_L")
-                                          || stateSupplier.get().isScoreCoral()
-                                          || stateSupplier.get() == SuperState.INTAKE_ALGAE_HIGH_LEFT
-                                          || stateSupplier.get() == SuperState.INTAKE_ALGAE_HIGH_RIGHT
-                                          || stateSupplier.get() == SuperState.INTAKE_ALGAE_LOW_LEFT
-                                          || stateSupplier.get() == SuperState.INTAKE_ALGAE_LOW_RIGHT)
+                                      && (Superstructure.getState().toString().startsWith("PRE_L")
```
spellingcat: `isScoreCoral` includes the pre states

src/main/java/frc/robot/camera/Camera.java
```diff
@@ -179,12 +178,12 @@ public void updateCamera(SwerveDrivePoseEstimator swerveEstimator) {
...
+                                          || Superstructure.getState().isScoreCoral()
```
spellingcat: static still feels like the wrong way to solve this but

spellingcat (approved):
i can just fix this one for you dw. i'm going to merge this and then merge the autoaim stuff after and that should be fine?

src/main/java/frc/robot/camera/Camera.java
SCool62: Ok.

SCool62:
Will merge #14 instead.

PR #11, opened 2025-10-04
Mttd superstructure

spellingcat:
main things:
- copy over pre-chezy updates to the superstructure (eg removing the list of transitions)
- update states to match the new offseason bot
- move requests fully inside the superstructure
- include end effector voltage in the rest of the state management to avoid the unpleasant case at chezy
- improve readability by moving various autoaim checks into their own methods

sans auto requests + scoring on both sides (i need to ruminate on these more) and the actual subsystems. i'm mostly looking for feedback on the state management stuff (and any edge cases i might have missed ☹️)


src/main/java/frc/robot/elevator/ElevatorSubsystem.java
kevinclark: Rather than using Units.inchesToMeters on every construction, you could just apply it to the argument here and rename to extensionInches.

kevinclark: Can apply this to some of our other enums too

src/main/java/frc/robot/Superstructure.java
```diff
@@ -41,14 +41,14 @@ public enum SuperState {
 
     INTAKE_CORAL_GROUND(ElevatorState.IDLE, ArmState.IDLE, IntakeState.INTAKE_CORAL),
 
-    PRE_HANDOFF(ElevatorState.IDLE, ArmState.IDLE, IntakeState.PRE_HANDOFF),
-    HANDOFF(ElevatorState.IDLE, ArmState.HANDOFF, IntakeState.HANDOFF),
+    READY_CORAL_INTAKE(ElevatorState.IDLE, ArmState.IDLE, IntakeState.READY_CORAL_INTAKE),
+    PRE_HANDOFF(ElevatorState.HANDOFF, ArmState.HANDOFF, IntakeState.READY_CORAL_INTAKE),
+    HANDOFF(ElevatorState.HANDOFF, ArmState.HANDOFF, IntakeState.HANDOFF),
+    READY_CORAL_ARM(ElevatorState.IDLE, ArmState.READY_CORAL_ARM, IntakeState.IDLE),
```
kevinclark, line 46: As someone approaching the code, it'd be awesome to see a common explaining what each of these states means.

kevinclark (changes requested):
In concept the changes seem fine, but a few small tweaks would make it easier for the next person to approach this.

PR #14, opened 2025-10-13
Autoaim commands

SCool62:


SCool62:
Not all autoaims have been tested, but the underlying translateToPose method works.

src/main/java/frc/robot/Robot.java
```diff
@@ -374,7 +408,7 @@ private void addAutos() {
   @Override
   public void robotPeriodic() {
     CommandScheduler.getInstance().run();
-    superstructure.periodic();
+    // superstructure.periodic();
```
spellingcat: why is this commented out?

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -333,95 +350,204 @@ public Command stop() {
...
+  /**
+   * Autoaligns to the given pose, stopping when it's within the passed-in tolerance
+   *
+   * @param target the target pose
+   * @param xToleranceMeters the allowed x-direction translational tolerance
+   * @param yToleranceMeters the allowed y-direction translational tolerance
+   * @param headingToleranceRadians the allowed heading tolerance
+   * @return a Command driving to the target pose
+   */
+  public Command translateToPoseWithTolerance(
```
spellingcat: idt this is used

src/main/java/frc/robot/utils/AutoAim.java
```diff
@@ -35,363 +38,62 @@ public class AutoAim {
...
-        && MathUtil.isNear(0.0, diff.getRotation().getRadians(), rotationTol);
+  static final ProfiledPIDController VX_CONTROLLER =
+      new ProfiledPIDController(
+          10.0, 0.01, 0.02, new Constraints(MAX_AUTOAIM_SPEED, MAX_AUTOAIM_ACCELERATION));
+  static final ProfiledPIDController VY_CONTROLLER =
+      new ProfiledPIDController(
+          10.0, 0.01, 0.02, new Constraints(MAX_AUTOAIM_SPEED, MAX_AUTOAIM_ACCELERATION));
+  static final ProfiledPIDController HEADING_CONTROLLER =
+      new ProfiledPIDController(
+          6.0, 0.0, 0.0, new Constraints(MAX_ANGULAR_SPEED, MAX_ANGULAR_ACCELERATION));
```
spellingcat, line 44: you'll need to enable continuous input so it wraps around (do negative pi to pi)

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
+   *
+   * @param target the target pose
+   * @param speedsModifier Field relative speeds to be added onto the PID calculated speeds. i.e.
+   *     driver requested speeds to drive along a line.
+   * @return a Command driving to the target
+   */
+  public Command translateToPose(Supplier<Pose2d> target, Supplier<ChassisSpeeds> speedsModifier) {
+    return Commands.runOnce(() -> AutoAim.resetPIDs(getPose(), getVelocityFieldRelative()))
+        .andThen(
+            driveClosedLoopFieldRelative(
```
spellingcat, line 428: is this velocity?

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
+                () -> {
+                  Logger.recordOutput("AutoAim/TargetPose", target.get());
+                  return AutoAim.calculateSpeeds(getPose(), target.get())
```
spellingcat, line 431: does get pose need to be a supplier

src/main/java/frc/robot/utils/AutoAim.java
```diff
@@ -35,363 +38,62 @@ public class AutoAim {
...
-    final var diff = pose.minus(pose2);
-    return MathUtil.isNear(
-            0.0, Math.hypot(diff.getX(), diff.getY()), AutoAim.TRANSLATION_TOLERANCE_METERS)
-        && MathUtil.isNear(
-            0.0, diff.getRotation().getRadians(), AutoAim.ROTATION_TOLERANCE_RADIANS);
+  public static ChassisSpeeds calculateSpeeds(
+      Pose2d robotPose, Pose2d target, Constraints constraints) {
+    VX_CONTROLLER.setConstraints(constraints);
+    VY_CONTROLLER.setConstraints(constraints);
+    // Should heading controller be set here as well? Or just translational?
```
spellingcat: maybe just change this to let you set separate constraints for each controller and then we can overload with more if we actually need them

src/main/java/frc/robot/utils/AutoAim.java
```diff
@@ -35,363 +38,62 @@ public class AutoAim {
...
+      new ProfiledPIDController(
+          10.0, 0.01, 0.02, new Constraints(MAX_AUTOAIM_SPEED, MAX_AUTOAIM_ACCELERATION));
+  static final ProfiledPIDController VY_CONTROLLER =
+      new ProfiledPIDController(
+          10.0, 0.01, 0.02, new Constraints(MAX_AUTOAIM_SPEED, MAX_AUTOAIM_ACCELERATION));
+  static final ProfiledPIDController HEADING_CONTROLLER =
+      new ProfiledPIDController(
+          6.0, 0.0, 0.0, new Constraints(MAX_ANGULAR_SPEED, MAX_ANGULAR_ACCELERATION));
+
+  public static void resetPIDs(Pose2d robotPos, ChassisSpeeds robotVelocityFieldRelative) {
```
spellingcat:
```suggestion
  public static void resetPIDControllers(Pose2d robotPos, ChassisSpeeds robotVelocityFieldRelative) {
```

src/main/java/frc/robot/swerve/SwerveSubsystem.java
spellingcat: i guess the better way of asking that question is is this equivalent to `driveVelocityFieldRelative` from last yr

src/main/java/frc/robot/utils/AutoAim.java
```diff
@@ -35,363 +38,62 @@ public class AutoAim {
...
-    final var diff = pose.minus(AlgaeIntakeTargets.getClosestTargetPose(pose));
-    return MathUtil.isNear(
-            0.0, Math.hypot(diff.getX(), diff.getY()), AutoAim.TRANSLATION_TOLERANCE_METERS)
-        && MathUtil.isNear(
-            0.0, diff.getRotation().getRadians(), AutoAim.ROTATION_TOLERANCE_RADIANS);
+  public static ChassisSpeeds calculateSpeeds(Pose2d robotPose, Pose2d target) {
+    VX_CONTROLLER.setConstraints(DEFAULT_TRANSLATIONAL_CONSTRAINTS);
+    VY_CONTROLLER.setConstraints(DEFAULT_TRANSLATIONAL_CONSTRAINTS);
+    HEADING_CONTROLLER.setConstraints(DEFAULT_ANGULAR_CONSTRAINTS);
+    return new ChassisSpeeds(
```
spellingcat: i think you'd have to do a check for if we're close enough to the target pose, and if so return an empty chassis speeds object

spellingcat: i think that's needed for the speedsmodifier stuff back in swerve subsystem (presumably this is to keep some element of the driver's initial velocity) i'd like to hear your rationale though if this was intentional

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
-    return Commands.none();
+  /**
+   * Autoailgns to the given pose
+   *
+   * @param target the target pose
+   * @param speedsModifier Field relative speeds to be added onto the PID calculated speeds. i.e.
+   *     driver requested speeds to drive along a line.
+   * @return a Command driving to the target
+   */
+  public Command translateToPose(Supplier<Pose2d> target, Supplier<ChassisSpeeds> speedsModifier) {
```
spellingcat: i'd make this command private. since it doesn't end, i don't think we'd ever want to call this outside of translateToPoseWithTolerance

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
+                }));
+  }
+
+  /**
+   * Autoaligns to the given pose
+   *
+   * @param target the target pose
+   * @return a Command driving to the target
+   */
+  public Command translateToPose(Supplier<Pose2d> target) {
```
spellingcat: also make this private

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
         });
   }
```
spellingcat, line 392: i can't seem to comment on it directly but i think L387 maybe should be a runOnce?

src/main/java/frc/robot/swerve/SwerveSubsystem.java
spellingcat: nvm i forgot autos exist and that we sometimes want it to end for other reasons. add something in the comment tho- "must manually define end condition" or something

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
+      double headingToleranceRadians) {
+    return translateToPose(target)
+        .until(
+            () ->
+                MathUtil.isNear(target.get().getX(), getPose().getX(), xToleranceMeters)
+                    && MathUtil.isNear(target.get().getY(), getPose().getY(), yToleranceMeters)
+                    && MathUtil.isNear(
+                        target.get().getRotation().getRadians(),
+                        getPose().getRotation().getRadians(),
+                        headingToleranceRadians));
```
spellingcat: could just call `isInTolerance` here

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
+  }
+
+  public boolean isInAutoAimTolerance(Pose2d target) {
+    return isInTolerance(
+        target, AutoAim.TRANSLATION_TOLERANCE_METERS, AutoAim.ROTATION_TOLERANCE_RADIANS);
+  }
+
+  public boolean isInTolerance(
+      Pose2d target, double translationalToleranceMeters, double angularToleranceRadians) {
```
spellingcat, line 461: i have a hard time envisioning a scenario where you have a different tolerance for one axis than another but if you're passing in both x and y into `translateToPoseWithTolerance` maybe make this match that also

spellingcat: ah, never mind

src/main/java/frc/robot/swerve/SwerveSubsystem.java
spellingcat: i have a hard time envisioning a scenario where you have a different tolerance for one axis than another so maybe just put in the translation tolerance instead of x and y individually

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
+    Transform2d diff = getPose().minus(target);
+    return MathUtil.isNear(0.0, Math.hypot(diff.getX(), diff.getY()), translationalToleranceMeters)
+        && MathUtil.isNear(
+            target.getRotation().getRadians(),
+            getPose().getRotation().getRadians(),
+            angularToleranceRadians);
+  }
+
+  public Command autoAimToL1(double vxModifier, double vyModifier) {
```
spellingcat: these should be `DoubleSupplier`s

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
+          Rectangle2d nearestLine = FieldUtils.L1Targets.getNearestLine(getPose());
+          return new Pose2d(
+              nearestLine.nearest(getPose().getTranslation()), nearestLine.getRotation());
+        },
+        () -> {
+          ChassisSpeeds speedsModifierRobotRelative =
+              ChassisSpeeds.fromFieldRelativeSpeeds(vxModifier, vyModifier, 0.0, getRotation());
+          // Kill all front-back requested velocity (because we want the robot to strafe)
+          speedsModifierRobotRelative.vxMetersPerSecond = 0.0;
+          return ChassisSpeeds.fromRobotRelativeSpeeds(speedsModifierRobotRelative, getRotation());
```
spellingcat: it appears there was an inexplicable `.times(1.5));` after this, do you remember why that was there

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
-    return true;
+    return isInAutoAimTolerance(
+        new Pose2d(
+            L1Targets.getNearestLine(getPose()).nearest(getPose().getTranslation()),
+            L1Targets.getNearestLine(getPose()).getRotation()));
   }
 
   public boolean isNearL1Reef() {
     // TODO
     // Not sure why this is different from near l1??
```
spellingcat, line 493: `nearL1` checks if we're close enough to actually score L1 (with kelpie). `isNearL1Reef` checks if we're close enough to the reef such that if we were to retract the (kelpie) elevator, the end effector would catch on the trough. probably not gonna be an issue with the madtown bot so just leave it for now

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
-    // TODO
-    return true;
+    return getPose()
+            .getTranslation()
+            .minus(
+                DriverStation.getAlliance().orElse(Alliance.Blue) == Alliance.Blue
+                    ? AutoAim.BLUE_REEF_CENTER
+                    : AutoAim.RED_REEF_CENTER)
+            .getNorm()
+        < 3.25;
```
spellingcat:
```suggestion
        < toleranceMeters;
```

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
   }
 
   public Command autoAimToL23() {
-    // TODO
-    return Commands.none();
+    return translateToPose(() -> FieldUtils.CoralTargets.getClosestTargetL23(getPose()));
```
spellingcat: add check for left vs right pole

spellingcat: and for that you'll need to get the velocity too

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
   }
 
   public boolean nearL23() {
-    // TODO
-    return true;
+    return isInAutoAimTolerance(FieldUtils.CoralTargets.getClosestTargetL23(getPose()));
   }
 
   public Command autoAimToL4() {
```
spellingcat: same thing about handed targets here

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
+    return translateToPose(() -> FieldUtils.CoralTargets.getClosestTargetL4(getPose()));
   }
 
   public boolean nearL4() {
-    // TODO
-    return true;
+    return isInAutoAimTolerance(FieldUtils.CoralTargets.getClosestTargetL4(getPose()));
   }
 
   public Command autoAimToOffsetAlgae() {
```
spellingcat:
```suggestion
  public Command autoAimToOffsetAlgaePose() {
```
or location or something

src/main/java/frc/robot/utils/AutoAim.java
```diff
@@ -35,363 +38,62 @@ public class AutoAim {
...
+        VX_CONTROLLER.calculate(robotPose.getX(), target.getX())
+            + VX_CONTROLLER.getSetpoint().velocity,
+        VY_CONTROLLER.calculate(robotPose.getY(), target.getY())
+            + VY_CONTROLLER.getSetpoint().velocity,
+        HEADING_CONTROLLER.calculate(
+                robotPose.getRotation().getRadians(), robotPose.getRotation().getRadians())
```
spellingcat:
```suggestion
                robotPose.getRotation().getRadians(), target.getRotation().getRadians())
```

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
-    return true;
+    return isInAutoAimTolerance(
+        FieldUtils.AlgaeIntakeTargets.getOffsetLocation(
+            FieldUtils.AlgaeIntakeTargets.getClosestTargetPose(getPose())));
   }
 
   public Command approachAlgae() {
-    // TODO
-    return Commands.none();
+    return translateToPose(() -> FieldUtils.AlgaeIntakeTargets.getClosestTargetPose(getPose()));
```
spellingcat: probably change this to translateToPose with the speeds modifier, since i think it's coming in with 1 m/s for vx (and also negates the output of the y controller - might have to do with it being robot relative)

spellingcat: also maybe rename it to like autoaim to algae or something. approach is kinda weird

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
   public boolean nearAlgaeIntakePose() {
-    // TODO
-    return true;
+    return isInAutoAimTolerance(FieldUtils.AlgaeIntakeTargets.getClosestTargetPose(getPose()));
   }
 
   public Command autoAimToProcessor() {
-    // TODO
-    return Commands.none();
+    return translateToPose(() -> getPose().nearest(AutoAim.PROCESSOR_POSES));
```
spellingcat: i think this was a `autoAimWithIntermediatePose` originally - still should be necessary because we don't want the arm + algae descending onto the wall if it's too close too early

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
   }
 
   public boolean nearProcessor() {
```
spellingcat, line 584: perchance also make a `nearProcessorIntermediate` or something (i'm bad at naming things)

spellingcat: and also add the autoaim with intermediate pose method

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
+                  // Are we going to need a seperate one for l4??
+                  return AutoAim.calculateSpeeds(
+                      getPose(),
+                      FieldUtils.CoralTargets.getClosestTargetL23(pose.get()),
+                      new Constraints(1.5, 1.0));
+                }));
+  }
+
+  public boolean isNearPoseAuto(Pose2d target) {
+    return isInTolerance(target, lastOdometryUpdateTimestamp, lastOdometryUpdateTimestamp);
```
spellingcat: me when autocomplete

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
-    return true;
-  }
-
-  public Command autoAimToBarge() {
-    // TODO
-    return Commands.none();
+    return isInAutoAimTolerance(getPose().nearest(AutoAim.PROCESSOR_POSES));
+  }
+
+  public Command autoAimToBarge(DoubleSupplier yModifier) {
```
spellingcat:
```suggestion
  public Command autoAimToBarge(DoubleSupplier vyModifier) {
```

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
+            AutoAim.getClosestBargeXCoord(getPose()),
+            getPose().getX(),
+            AutoAim.TRANSLATION_TOLERANCE_METERS)
+        && MathUtil.isNear(
+            AutoAim.getClosestBargeRotation(getPose()).getRadians(),
+            getPose().getRotation().getRadians(),
+            AutoAim.ROTATION_TOLERANCE_RADIANS);
   }
 
   public Command autoAimAuto(Supplier<Pose2d> pose) {
```
spellingcat: is this actually different from autoAimToL23()/translateToPose() ? maybe just add an overload that takes in the vel/accel constraints

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -379,136 +392,204 @@ public Command stop() {
...
-  }
-
-  public boolean isNearPoseAuto(Supplier<Pose2d> pose) {
-    // TODO
-    return true;
+    return Commands.runOnce(() -> AutoAim.resetPIDs(getPose(), getVelocityFieldRelative()))
+        .andThen(
+            driveClosedLoopFieldRelative(
+                () -> {
+                  // Are we going to need a seperate one for l4??
```
spellingcat: i suppose you could pass in a coral score target and do a little selection logic of which closest target method you want to call based on that

src/main/java/frc/robot/swerve/SwerveSubsystem.java
spellingcat: this probably doesn't need to exist, i think it's just the same as `isInAutoAimTolerance`

src/main/java/frc/robot/utils/AutoAim.java
```diff
@@ -1,21 +1,24 @@
...
 import edu.wpi.first.math.util.Units;
-import frc.robot.utils.FieldUtils.AlgaeIntakeTargets;
-import frc.robot.utils.FieldUtils.CoralTargets;
 import java.util.List;
 
 public class AutoAim {
   static final double MAX_ANGULAR_SPEED = 10.0;
   static final double MAX_ANGULAR_ACCELERATION = 10.0;
   static final double MAX_AUTOAIM_SPEED = 3.0;
   static final double MAX_AUTOAIM_ACCELERATION = 4.0;
```
spellingcat:
```suggestion
  static final double MAX_TRANSLATIONAL_SPEED = 3.0;
  static final double MAX_TRANSLATIONAL_ACCELERATION = 4.0;
```

src/main/java/frc/robot/utils/AutoAim.java
spellingcat: you're gonna hate me, but i personally think the field constants like these should go into `FieldUtils` instead

src/main/java/frc/robot/utils/AutoAim.java
```diff
@@ -35,363 +38,62 @@ public class AutoAim {
...
-            0.0, Math.hypot(diff.getX(), diff.getY()), AutoAim.TRANSLATION_TOLERANCE_METERS)
-        && MathUtil.isNear(
-            0.0, diff.getRotation().getRadians(), AutoAim.ROTATION_TOLERANCE_RADIANS);
-  }
-
-  public static boolean isInToleranceCoral(Pose2d pose, double translationTol, double rotationTol) {
-    final var diff = pose.minus(CoralTargets.getClosestTarget(pose));
-    return MathUtil.isNear(0.0, Math.hypot(diff.getX(), diff.getY()), translationTol)
-        && MathUtil.isNear(0.0, diff.getRotation().getRadians(), rotationTol);
+  static final ProfiledPIDController VX_CONTROLLER =
```
spellingcat, line 30: maybe specify that the v stands for velocity in a comment

src/main/java/frc/robot/utils/AutoAim.java
```diff
@@ -35,363 +38,62 @@ public class AutoAim {
...
   }
 
-  public static boolean isInTolerance(Pose2d pose1, Pose2d pose2, ChassisSpeeds speeds) {
-    return isInTolerance(
-        pose1,
-        pose2,
-        speeds,
-        AutoAim.TRANSLATION_TOLERANCE_METERS,
-        AutoAim.ROTATION_TOLERANCE_RADIANS);
+  public static double getClosestBargeXCoord(Pose2d pose) {
```
spellingcat, line 90: supplierify?

src/main/java/frc/robot/utils/AutoAim.java
```diff
@@ -35,363 +38,62 @@ public class AutoAim {
...
-      double rotationTolerance) {
-    final var diff = pose1.minus(pose2);
-    return MathUtil.isNear(0.0, Math.hypot(diff.getX(), diff.getY()), translationTolerance)
-        && MathUtil.isNear(0.0, diff.getRotation().getRadians(), rotationTolerance)
-        && MathUtil.isNear(
-            0,
-            Math.hypot(speeds.vxMetersPerSecond, speeds.vyMetersPerSecond),
-            VELOCITY_TOLERANCE_METERSPERSECOND)
-        && MathUtil.isNear(0.0, speeds.omegaRadiansPerSecond, 3.0);
+  public static Rotation2d getClosestBargeRotation(Pose2d pose) {
```
spellingcat, line 97: supplierify?

src/main/java/frc/robot/utils/AutoAim.java
```diff
@@ -35,363 +38,62 @@ public class AutoAim {
...
+    return (Math.abs(pose.getX() - AutoAim.BLUE_NET_X) > Math.abs(pose.getX() - AutoAim.RED_NET_X)
+            ? Rotation2d.kZero
+            : Rotation2d.k180deg)
+        .plus(Rotation2d.fromDegrees(20.0));
```
spellingcat, line 103: add a TODO on this cause this will change

src/main/java/frc/robot/utils/FieldUtils.java
```diff
@@ -204,23 +204,39 @@ private CoralTargets(Pose2d location, boolean leftHanded) {
...
+          new Transform2d(
+              0.291
+                  + (SwerveSubsystem.SWERVE_CONSTANTS.getBumperLength() / 2)
+                  + Units.inchesToMeters(4.7),
+              Units.inchesToMeters(-7.879),
+              Rotation2d.fromDegrees(180)));
+    }
+
     public static Pose2d getBranchLocation(Pose2d transformed) {
       // 0.248 for trough
```
spellingcat: do these need to be public?

src/main/java/frc/robot/utils/FieldUtils.java
```diff
@@ -204,23 +204,39 @@ private CoralTargets(Pose2d location, boolean leftHanded) {
...
+      return original.transformBy(
+          new Transform2d(
+              0.291
+                  + (SwerveSubsystem.SWERVE_CONSTANTS.getBumperLength() / 2)
+                  + Units.inchesToMeters(4.7),
+              Units.inchesToMeters(-7.879),
+              Rotation2d.fromDegrees(180)));
+    }
+
     public static Pose2d getBranchLocation(Pose2d transformed) {
```
spellingcat: don't need

src/main/java/frc/robot/utils/FieldUtils.java
spellingcat: `leftHanded`

src/main/java/frc/robot/utils/FieldUtils.java
```diff
@@ -243,7 +263,7 @@ public static Pose2d getHandedClosestTarget(Pose2d pose, boolean leftHandeed) {
               .filter((target) -> target.leftHanded == leftHandeed)
               .map(
                   (CoralTargets targets) -> {
-                    return CoralTargets.getRobotTargetLocation(targets.location);
+                    return CoralTargets.getRobotTargetLocationL23(targets.location);
                   })
               .toList());
```
spellingcat, line 272: if you're feeling up for it later, maybe also add the stack coral targets. although i don't imagine these would see much use in teleop so probably wait on that actually

spellingcat: anyway if/when you do i think making those translation2ds (pose2ds without the rotation) probably makes the most sense? dw about it tho

spellingcat (changes requested):
wow this one was a doozy. genuinely thank you for doing this, and so cleanly too lol. minor bugs

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: yes

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: getPose() already is a supplier. the java suppliers are just a way of passing functions that take no inputs and return outputs (like getters) as values. Everytime the command scheduler executes this block, getPose() will be called. At least that's my understanding

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: ig so.

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: We aren't using this method anyways so i've deleted it. If we end up needing it its easy enough to add back.

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: Yeah. forgot about that. I think it was to give the driver more control (because otherwise the autoaim would fight the driver)

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: lol

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: Auto aim auto takes in the trajectory end pose so it can calculate the closest target, instead of just using the robot pose.

src/main/java/frc/robot/utils/AutoAim.java
SCool62: Shouldn't be necessary bc its just a run-of-the mill function, not a command. When it gets called inside of a supplier (like in autoAimToBarge), it will get called every scheduler run with the updated pose.

src/main/java/frc/robot/utils/AutoAim.java
SCool62: Same as last

src/main/java/frc/robot/utils/FieldUtils.java
SCool62: It's unused. should I remove?

SCool62: lol nvm

src/main/java/frc/robot/utils/FieldUtils.java
SCool62: lol

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: I assume this will work the same with left vs right bumpers?

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: idk I think approach is good because that's what it's doing.

src/main/java/frc/robot/camera/Camera.java
```diff
@@ -215,7 +218,6 @@ public void updateCamera(SwerveDrivePoseEstimator swerveEstimator) {
                                   || stateSupplier.get() == SuperState.PRE_BARGE_RIGHT
                               ? 0.5
                               : 1.0));
-              // the sussifier
             });
```
spellingcat: ik it's stupid but this genuinely helped me find it lmao

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -531,23 +532,46 @@ public boolean nearIntakeAlgaeOffsetPose() {
...
+          AutoAim.resetPIDControllers(getPose(), getVelocityFieldRelative());
+          ChassisSpeeds calculatedSpeedsRobotRelative =
+              ChassisSpeeds.fromFieldRelativeSpeeds(
+                  AutoAim.calculateSpeeds(
+                      getPose(), FieldUtils.AlgaeIntakeTargets.getClosestTargetPose(getPose())),
+                  getRotation());
+          calculatedSpeedsRobotRelative.vxMetersPerSecond =
+              isInAutoAimTolerance(FieldUtils.AlgaeIntakeTargets.getClosestTargetPose(getPose()))
+                  ? AutoAim.ALGAE_APPROACH_SPEED_METERS_PER_SECOND
+                  : 0.0;
```
spellingcat: other way around- if it _is_ in tolerance it should be 0

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -371,144 +386,235 @@ public Command driveOpenLoopFieldRelative(Supplier<ChassisSpeeds> speeds) {
...
+          // IDK why we have to do this but it was in reefscape so...
+          calculatedSpeedsRobotRelative.vyMetersPerSecond =
+              -calculatedSpeedsRobotRelative.vyMetersPerSecond;
```
spellingcat:
```suggestion
                    calculatedSpeedsRobotRelative.vyMetersPerSecond *= -1;

```
minus signs are hard to read imo

src/main/java/frc/robot/swerve/SwerveSubsystem.java
spellingcat: kk. in that case i would rename the parameter here to `endingPathPose` or similar

src/main/java/frc/robot/swerve/SwerveSubsystem.java
```diff
@@ -371,144 +386,235 @@ public Command driveOpenLoopFieldRelative(Supplier<ChassisSpeeds> speeds) {
...
-    return true;
+    return Commands.runOnce(
+            () -> AutoAim.resetPIDControllers(getPose(), getVelocityFieldRelative()))
+        .andThen(
+            driveClosedLoopFieldRelative(
+                () -> {
+                  return AutoAim.calculateSpeeds(
+                      getPose(),
+                      // TODO TUNE THIS. IT SHOULD BE LIKE HALFWAY BETWEEN l2 and l4
+                      isNearReef(Units.inchesToMeters(42))
```
spellingcat: maybe just check what `Robot.getCoralScoreTarget()` is instead

src/main/java/frc/robot/utils/AutoAim.java
spellingcat: in case you haven't noticed i'm bad at figuring out where to use suppliers 💀

src/main/java/frc/robot/swerve/SwerveSubsystem.java
spellingcat: on second thought, the contents of the method still feel redundant. can you just do translateToPose(L2/3 ? get closest l2/3 target to endingPathPose : get closest l4 target to endingPathPose)

src/main/java/frc/robot/Autos.java
```diff
@@ -189,7 +189,7 @@ public Command runPathThenIntakeCoralGround(Path path, AutoRoutine routine) {
   public Command scoreCoralInAuto(Supplier<Pose2d> trajEndPose) {
     return Commands.sequence(
             Commands.waitUntil(
-                new Trigger(() -> swerve.isNearPoseAuto(trajEndPose))
+                new Trigger(() -> swerve.isInAutoAimTolerance(trajEndPose.get()))
```
spellingcat: slight distinction here- we don't want to be close to the end pose, we want to be close to the _scoring position_ closest to the end pose

spellingcat (changes requested):
small stuff

src/main/java/frc/robot/camera/Camera.java
SCool62: Do you want me to put it back?

src/main/java/frc/robot/swerve/SwerveSubsystem.java
SCool62: lol

spellingcat (approved):

PR #16, opened 2025-10-21
Correct elevator motor inversion value

SCool62:
Apparently, clockwise vs counterclockwise is from the perspective of looking at the motor.


spellingcat (approved):
i'll take your word for it. i suppose we will know soon enough lol

PR #18, opened 2025-11-04
Autoaim fix

SCool62:


src/main/java/frc/robot/swerve/constants/OffseasonBotSwerveConstants.java
```diff
@@ -233,7 +233,7 @@ public CameraConstants[] getCameraConstants() {
         new CameraConstants(
             "Right_Elevator",
             new Transform3d(
-                new Translation3d(0.287, -0.139, 0.821),
+                new Translation3d(-0.139, -0.211, 0.798),
```
spellingcat, line 304: where are you getting these from?

spellingcat: the vision transforms seemed largely correct irl yesterday

spellingcat (commented):
reviewing on phone rn so may have missed something. will test irl once mech is done

PR #19, opened 2025-11-04
attempting to add zeroing sequence for arm, intake, and elevator (not tested)

vivi-o:


src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -44,6 +45,7 @@ public class ArmSubsystem extends RollerPivotSubsystem {
   public static final double CORAL_CURRENT_THRESHOLD = 30.0;
   public static final double TOLERANCE_DEGREES = 10.0;
   public static final double VERTICAL_OFFSET_METERS = Units.inchesToMeters(12.0);
+  public static final double SAFE_ZEROING_ANGLE = 120.0; // idk
```
spellingcat: recap of what ed's suggestion just now:
add some button binding where when (probably the operator) they press it the arm switches to voltage control and they just sort of jog it around until DT says visually it seems correct,t hen the rest of the zeroing stuff can happen

this has to override the default state position command Somehow.

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -205,4 +209,16 @@ public Command rezeroFromEncoder() {
   public Command rezeroAgainstRightBumper() {
     return zeroPivot(() -> Rotation2d.fromDegrees(ZEROING_ANGLE));
   }
+
+  public Command runCurrentZeroing() {
```
spellingcat, line 249: add something after the stuff has zeroed to go back to either idle, ready coral arm, ready coral intake, or ready algae depending on if we have a piece or not

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -205,4 +209,16 @@ public Command rezeroFromEncoder() {
...
+    return this.run(() -> setPivotVoltage(() -> -3.0))
```
spellingcat, line 250: this should be positive i think

src/main/java/frc/robot/rollerpivot/RollerPivotSubsystem.java
```diff
@@ -39,7 +39,7 @@ protected void runRollerVelocity(double velocityRPS) {
     rollerIO.setRollerVelocity(velocityRPS);
   }
 
-  protected void setPivotAngle(Supplier<Rotation2d> target) {
+  public void setPivotAngle(Supplier<Rotation2d> target) {
```
spellingcat: change back to protected bc the driver is gonna manually jog

spellingcat: sam says you should unsupplier this

src/main/java/frc/robot/Robot.java
```diff
@@ -656,6 +656,20 @@ private void addControllerBindings() {
                             // : Rotation2d.kCCW_90deg)));
                             ? Rotation2d.kZero
                             : Rotation2d.k180deg)));
+
+    driver
+        .start()
+        .onTrue(
```
spellingcat, line 861: i think while true might be more correct?i forgot what we did for kelpie tho so check against that

spellingcat (changes requested):

src/main/java/frc/robot/arm/ArmSubsystem.java
vivi-o: Just to double check, right now it will go back to the state it was in, but instead we want it to just go to ready coral/algae or idle instead?

spellingcat: yes

src/main/java/frc/robot/Robot.java
```diff
@@ -659,16 +667,12 @@ private void addControllerBindings() {
...
-                // doent slam it againt something else
-                Commands.runOnce(
-                    () -> arm.setPivotAngle(() -> Rotation2d.fromDegrees(arm.SAFE_ZEROING_ANGLE))),
-                elevator.runCurrentZeroing(),
-                arm.runCurrentZeroing()
-                // () -> arm.rezeroFromEncoder(),
+                intake.runCurrentZeroing(), elevator.runCurrentZeroing(), arm.rezeroFromEncoder()
+                // TODO add transition to specific states after zeroing
+                // only use if cancoder is cooked use:
+                // arm.runCurrentZeroing()
```
spellingcat: this looks good and we can test it today. another thing that would be good to add would be another trigger for it to run thru the zeroing sequence when we first enable

so in intake+elevator, add a boolean called like `hasZeroedSinceStartup` that's set to false originally. then add a new trigger that's like:

(pseudocode)
new Trigger(state == idle)
.and(() -> !elevator.hasZeroedSinceStartup || !intake.hasZeroedSinceStartup)
.and(() -> driver station is enabled)
.onTrue(
  Commands.sequence
    (intake.runCurrentZeroing(), 
    elevator.runCurrentZeroing(), 
    arm.rezeroFromEncoder(),
    elevator.hasZeroedSinceStartup = true
    intake.hasZeroedSinceStartup = true
))

spellingcat: also could you add @AutoLogOutput to each of those has zeroed since startup variables

src/main/java/frc/robot/Robot.java
```diff
@@ -644,6 +644,14 @@ private void addControllerBindings() {
     // operator.povRight().onTrue(Commands.runOnce(() -> leftHandedTarget = false));
 
     // heading reset
+    // figure out how to make sure it doesnt jump back to default make add into zeroing button
+    operator.povLeft().whileTrue(Commands.runOnce(() -> arm.setPivotVoltage(-3.0)));
+
+    operator.povRight().whileTrue(Commands.runOnce(() -> arm.setPivotVoltage(3.0)));
```
spellingcat: operator.povLeft().whileTrue(arm.setPivotVoltage(-3.0).withTimeout(0.05))).

setPivotVoltage is already a command and adding the timeout means it only jogs it for a little while. then also do that for right

src/main/java/frc/robot/Robot.java
```diff
@@ -644,6 +644,14 @@ private void addControllerBindings() {
     // operator.povRight().onTrue(Commands.runOnce(() -> leftHandedTarget = false));
 
     // heading reset
+    // figure out how to make sure it doesnt jump back to default make add into zeroing button
```
spellingcat: hmm yeah that's a good point. the play might be to add a sort of hold() method like:

public Command hold() {
  return Commands.sequence(
      setTargetAngle(() -> inputs.position).until(() -> true), this.run(() -> {}));
}

don't ask me how this works i copied it from kelpie. i think the this.run at the end "holds onto" the command so it doesn't end unless it's explicitly interrupted by smth else

so then do that and then like
Commands.runOnce(() -> arm.setPivotVoltage(-3.0)).andThen(hold().until(elevator is zeroed ));

you might need to make a new variable for the interrupt condition

i can explain more at lunch or something or we can discuss at some point sorry im typing this out right before i leave for school lmao

spellingcat (commented):
looking good i have a couple thoughts on more things we could add but i'll write them up later

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -198,6 +198,15 @@ public Command setStateAngleVelocity() {
     // return this.run(() -> runRollerVelocity(getState().velocityRPS.getAsDouble()));
   }
 
+  public Command hold() {
+    return Commands.sequence(
+        Commands.runOnce(() -> setPivotAngle(cancoderInputs.cancoderPositionRotations))
```
spellingcat: i'd change this to use the motor position, given that we're really only ever going to manually rezero the arm when the cancoder is down

spellingcat: it's ok if the motor position ends up being incorrect from real life- we just want it to stay where it thinks it is

src/main/java/frc/robot/Robot.java
```diff
@@ -643,14 +643,26 @@ private void addControllerBindings() {
...
 
-    // heading reset
-    // figure out how to make sure it doesnt jump back to default make add into zeroing button
-    operator.povLeft().whileTrue(Commands.runOnce(() -> arm.setPivotVoltage(-3.0)));
-
-    operator.povRight().whileTrue(Commands.runOnce(() -> arm.setPivotVoltage(3.0)));
+    // TODO add escape for hold??
+    operator
+        .povLeft()
+        .whileTrue(
```
spellingcat: perhaps change to onTrue actually since there's the timeout

src/main/java/frc/robot/Robot.java
```diff
@@ -643,14 +643,26 @@ private void addControllerBindings() {
...
     // operator.povRight().onTrue(Commands.runOnce(() -> leftHandedTarget = false));
 
-    // heading reset
-    // figure out how to make sure it doesnt jump back to default make add into zeroing button
-    operator.povLeft().whileTrue(Commands.runOnce(() -> arm.setPivotVoltage(-3.0)));
-
-    operator.povRight().whileTrue(Commands.runOnce(() -> arm.setPivotVoltage(3.0)));
+    // TODO add escape for hold??
+    operator
+        .povLeft()
```
spellingcat: ok wait sorry, i thought of a better way to do this. let's add a trigger called preZeroingReq or something like that and bind it to some unused button or whatever. then, the conditions for the jog starting is:

(preZeroingReq AND operator povleft) WHILE TRUE: set arm voltage with timeout AND THEN hold

this way when the prezeroingreq goes to false it'll stop being controlled by the jog.

spellingcat:
will figure this out tn with vivie over slack

spellingcat (changes requested):
this one thing- and then also triggerifying it which i messaged you about

src/main/java/frc/robot/Robot.java
```diff
@@ -78,6 +78,9 @@
 public class Robot extends LoggedRobot {
   public static final RobotType ROBOT_TYPE = Robot.isReal() ? RobotType.REAL : RobotType.SIM;
   public static final boolean TUNING_MODE = true;
+  public boolean preZeroingReq = false;
+  public boolean zeroingReq = false;
```
spellingcat: these should be `Trigger`s mapped to buttons on the controllers. this means you won't have to manually set the booleans or do the () -> every time

src/main/java/frc/robot/Robot.java
```diff
@@ -743,7 +746,6 @@ private void addControllerBindings() {
     // operator.povLeft().onTrue(Commands.runOnce(() -> leftHandedTarget = true));
     // operator.povRight().onTrue(Commands.runOnce(() -> leftHandedTarget = false));
 
-    // heading reset
```
spellingcat, line 782: maybe keep this comment

src/main/java/frc/robot/Robot.java
```diff
@@ -756,6 +758,54 @@ private void addControllerBindings() {
...
+    operator
+        .povDown()
+        .and(() -> preZeroingReq)
+        .whileTrue(Commands.parallel(arm.setPivotVoltage(-3.0)).withTimeout(0.05));
+
+    operator
+        .povUp()
+        .and(() -> preZeroingReq)
+        .whileTrue(Commands.parallel(arm.setPivotVoltage(3.0)).withTimeout(0.05));
+
```
spellingcat: also add one that's like, pre zeroing req AND no operator up AND no operator down: arm.hold()

spellingcat: that way operator can freeze it in place between jogs

src/main/java/frc/robot/Robot.java
```diff
@@ -756,6 +758,55 @@ private void addControllerBindings() {
...
+    // preZeroingReq
+    driver.start().onTrue(Commands.runOnce(() -> preZeroingReq = true));
```
spellingcat: in the pre zeroing req definition you can do like

Trigger preZeroingReq = driver.start();

src/main/java/frc/robot/Robot.java
```diff
@@ -756,6 +758,55 @@ private void addControllerBindings() {
...
+
+    new Trigger(
+            () ->
+                (preZeroingReq)
+                    && !operator.povUp().getAsBoolean()
+                    && !operator.povDown().getAsBoolean())
+        .whileTrue(arm.hold());
```
spellingcat, line 831: you don't need the until- the trigger includes the prezerongreq and zeroingreq so when either of those are false, the whole trigger is false and it stops

src/main/java/frc/robot/Robot.java
```diff
@@ -756,6 +758,55 @@ private void addControllerBindings() {
...
+        .povUp()
+        .and(() -> preZeroingReq)
+        .whileTrue(
+            Commands.parallel(
+                    Commands.runOnce(() -> preZeroingReq = false),
+                    Commands.runOnce(() -> zeroingReq = true),
+                    intake.runCurrentZeroing(),
+                    elevator.runCurrentZeroing(),
+                    arm.hold()
+                    // if cancoder is cooked it would use:
```
spellingcat: i would uncomment this just because i think the most likely situation we'll use this sequence in is if the cancoder is cooked

src/main/java/frc/robot/Robot.java
```diff
@@ -756,6 +758,55 @@ private void addControllerBindings() {
...
+                    // arm.runCurrentZeroing()
+                    )
+                .andThen(
+                    Commands.parallel(
```
spellingcat, line 845: i'm reviewing on my phone rn so i'm really sorry if im just missing something but i don't think this is needed because arm.runcurrentzeroing calls it when it ends

src/main/java/frc/robot/Superstructure.java
```diff
@@ -364,6 +364,29 @@ private void setSubstates() {
     climber.setState(state.climberState);
   }
 
+  // idk if this works right states confuse me
```
spellingcat: 👍👍👍

src/main/java/frc/robot/Superstructure.java
```diff
@@ -364,6 +364,29 @@ private void setSubstates() {
...
+  public Command transitionAfterZeroing() {
+    return Commands.runOnce(
+            () -> {
+              SuperState target;
+              // cant distinguish between coral and algae for intake so should that be fixed with
+              // the intake or what should i do here?
```
spellingcat: dw about it (we probably won't get algae working 💀)

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -218,4 +231,16 @@ public Command rezeroFromEncoder() {
   public Command rezeroAgainstRightBumper() {
     return zeroPivot(() -> Rotation2d.fromDegrees(ZEROING_ANGLE));
   }
+
+  public Command runCurrentZeroing() {
+    return this.run(() -> setPivotVoltage(3.0))
```
spellingcat: this should be negative because negative is to the right

src/main/java/frc/robot/rollerpivot/RollerPivotSubsystem.java
```diff
@@ -31,21 +30,21 @@ public RollerPivotSubsystem(RollerIO rollerIO, PivotIO pivotIO, String name) {
...
-    pivotIO.setMotorPosition(target.get());
+  protected void setPivotAngle(Rotation2d target) {
+    Logger.recordOutput(name + "/Pivot Setpoint", target);
+    pivotIO.setMotorPosition(target);
   }
 
-  public Command setPivotVoltage(DoubleSupplier volts) {
-    return this.run(() -> pivotIO.setMotorVoltage(volts.getAsDouble()));
+  public Command setPivotVoltage(double volts) {
+    return this.run(() -> pivotIO.setMotorVoltage(volts));
```
spellingcat: i think commands need to have suppliers (but normal methods don't)

spellingcat (changes requested):
super close, nice work :D

src/main/java/frc/robot/Superstructure.java
vivi-o: ;-;

src/main/java/frc/robot/Robot.java
spellingcat: oops sorry realized you did do that later down

src/main/java/frc/robot/Robot.java
vivi-o: what does this mean?

spellingcat: lol for some reason i can't find the line i was talking about anymore. reviewing on a phone is truly a terrible experience💀 i'll look at it when i get to my computer again

spellingcat (approved):

PR #20, opened 2025-11-06
Climber tuning

SCool62:


src/main/java/frc/robot/Robot.java
```diff
@@ -416,17 +425,17 @@ public Robot() {
         arm.rezeroAgainstRightBumper()
             .alongWith(Commands.print("dashboard rezero arm against bumper"))
             .ignoringDisable(true));
+    SmartDashboard.putData(
+        "Spool in climber (MANUAL STOP)",
+        Commands.parallel(climber.retract(), intake.setPivotVoltage(() -> -4.0)));
```
spellingcat: to be extra safe i'd do a Commands.parallel(set intake voltage, commands.wait(intake current spikes).andThen(climber retract)); or something like that

spellingcat: ig it wouldn't be parallel so whatever like deadline/racewith

src/main/java/frc/robot/Robot.java
```diff
@@ -368,10 +377,10 @@ public Robot() {
     PhoenixOdometryThread.getInstance().start();
 
     // Set default commands
-    elevator.setDefaultCommand(elevator.setStateExtension());
-    arm.setDefaultCommand(arm.setStateAngleVelocity());
-    intake.setDefaultCommand(intake.setStateAngleVelocity());
-    // climber.setDefaultCommand(climber.setStateAngleVoltage());
+    // elevator.setDefaultCommand(elevator.setStateExtension());
+    // arm.setDefaultCommand(arm.setStateAngleVelocity());
+    // intake.setDefaultCommand(intake.setStateAngleVelocity());
```
spellingcat: uncomment these (ik it was prob for testing)

src/main/java/frc/robot/climber/ClimberSubsystem.java
```diff
@@ -30,8 +30,8 @@ public enum ClimberState {
     IDLE(Rotation2d.fromDegrees(0), 0.0),
     // climbing
     // TODO climber stuff is very tbd rn
-    PRE_CLIMB(Rotation2d.fromDegrees(0), 0.0),
-    CLIMB(Rotation2d.fromDegrees(20), 0.0);
+    PRE_CLIMB(Rotation2d.fromRadians(17.93), -5.0),
+    CLIMB(Rotation2d.fromDegrees(0.0), 0.0);
```
spellingcat, line 34: i'm surprised that roller voltage is 0?

src/main/java/frc/robot/Robot.java
```diff
@@ -368,10 +377,10 @@ public Robot() {
...
+    climber.setDefaultCommand(climber.setStateAngleVoltage());
```
spellingcat, line 384: do me a favor and add in a comment that voltage instead of velocity is intentional here so we don't forget later

spellingcat (changes requested):
couple of things

src/main/java/frc/robot/climber/ClimberSubsystem.java
SCool62: For climb?

SCool62: The locking plates do all the work. The rollers are only for centering

spellingcat: oh real i forgot about that

src/main/java/frc/robot/rollerpivot/RollerPivotSubsystem.java
```diff
@@ -93,7 +93,7 @@ public void periodic() {
     rollerIO.updateInputs(rollerInputs);
     Logger.processInputs(name + "/Roller", rollerInputs);
 
-    currentFilterValue = currentFilter.calculate(rollerInputs.statorCurrentAmps);
+    currentFilterValue = currentFilter.calculate(pivotInputs.statorCurrentAmps);
```
spellingcat: this doesn't seem right- the arm uses the roller currentfiltervalue to determine the presence or absence of a game piece

spellingcat: i'd just add another method in intake subsystem specifically that looks at the pivot

spellingcat (changes requested):

src/main/java/frc/robot/rollerpivot/RollerPivotSubsystem.java
SCool62: BTW the zeroing method needs changing too then bc it uses the roller current filter value

spellingcat (approved):

PR #27, opened 2025-12-10
Angled handoff

spellingcat:


src/main/java/frc/robot/Robot.java
```diff
@@ -876,7 +882,8 @@ private void addControllerBindings() {
                 Commands.deadline(
                     arm.runCurrentZeroing(),
                     Commands.parallel(
-                        elevator.setVoltage(() -> -1.0), intake.setPivotVoltage(() -> -3.0))),
+                        elevator.setVoltage(() -> -1.0).repeatedly(),
+                        intake.setPivotVoltage(() -> -3.0).repeatedly())),
```
SCool62, line 886: Doesn't need to be repeatedly but idt it matters that much

src/main/java/frc/robot/arm/ArmSubsystem.java
```diff
@@ -206,24 +185,65 @@ public boolean isNearAngle(Rotation2d target) {
...
-    //     () -> getState().position.get(), () -> getState().velocityRPS.getAsDouble());
-    // return this.run(() -> runRollerVelocity(getState().velocityRPS.getAsDouble()));
     return this.run(
         () -> {
-          if (Superstructure.getState() == SuperState.CLIMB
-              || Superstructure.getState() == SuperState.PRE_CLIMB) {
-            pivotIO.setMotorVoltage(-3.0);
-          } else {
-            pivotIO.setMotorPosition(state.getAngle());
+          switch (Superstructure.getState()) {
```
SCool62, line 190: This switch seems like prime bug creation location.

SCool62 (approved):

src/main/java/frc/robot/arm/ArmSubsystem.java
spellingcat: i know it's terrible

src/main/java/frc/robot/Robot.java
spellingcat: i'm p sure it does because otherwise it won't hold itself

PR #30, opened 2025-12-12
Sysid

SCool62:
The robot code facing side of SysId seems pretty simple. We should run some successful sysids on the robot before merging. I'll also make a SysId article in Highlanders training documenting the process of creating, running, and analyzing a sysid.


SCool62:
Tested elevator sysid. After a little fiddling in the sysid tool, I was able to get calculated feedforward gains that pretty accurately hit a velocity setpoint on feedforward control only (and were better than the gains we found by manual tuning). Haven't tested arm or roller sysid yet but should be a similar process.

PR #32, opened 2026-01-09
use ctre sim for elevator

spellingcat:
this doesn't actually need to be merged but i'm tagging y'all for review


src/main/java/frc/robot/elevator/ElevatorIOCTRESim.java
```diff
@@ -0,0 +1,89 @@
...
+      new ProfiledPIDController(110.0, 0.0, 0.0, new Constraints(5.0, 10.0));
+  private final ElevatorFeedforward ff =
+      new ElevatorFeedforward(
+          0.24,
+          0.56,
+          (DCMotor.getKrakenX60Foc(1).KvRadPerSecPerVolt
+                  * (ElevatorSubsystem.SPROCKET_DIAMETER_METERS / 2))
+              / ElevatorSubsystem.GEAR_RATIO);
+
+  // automatically calls the superclass constructor
```
SCool62: It looks like in the docs there's a couple of config lines that are supposed to happen in simulationInit() (like setting the motor type etc.). Maybe those should go in the constructor?

src/main/java/frc/robot/elevator/ElevatorIOCTRESim.java
```diff
@@ -0,0 +1,89 @@
...
+  @Override
+  public void updateInputs(ElevatorIOInputs inputs) {
+    if (DriverStation.isDisabled()) {
+      stop();
+    }
+    // set the supply voltage of the motor
+    leaderSim.setSupplyVoltage(RobotController.getBatteryVoltage());
+
+    // calculate pid for the physics sim
+    volts =
```
SCool62: I had thought that a main advantage of this is instead of using the wpilib controllers, the controls use the talonfx config as they would on robot, and the physics sim gets the voltage from the motor sim state. lmk if i'm mistaken tho

src/main/java/frc/robot/elevator/ElevatorIOCTRESim.java
```diff
@@ -0,0 +1,89 @@
...
+    physicsSim.setInputVoltage(MathUtil.clamp(volts, -12.0, 12.0));
+    // physicsSim.setInputVoltage((-1) * leaderSim.getMotorVoltageMeasure().in(Volts));
+    physicsSim.update(0.020); // assume 20 ms loop time
+
+    // update the ctre motor sim to match the wpilib physics sim
+    // note that "set" does not command the motor to do anything
+    // it just updates the value the ctre sim has stored
+
+    // convert back to rotor position because it doesn't let you set linear position directly
+    leaderSim.setRawRotorPosition(
```
SCool62: So strange they don't let you set actual mech position. CTRE is stupid sometimes ig.

src/main/java/frc/robot/elevator/ElevatorIOReal.java
```diff
@@ -38,6 +38,9 @@ public class ElevatorIOReal implements ElevatorIO {
   private DynamicMotionMagicVoltage motionMagicVoltage;
   private TorqueCurrentFOC torqueCurrent = new TorqueCurrentFOC(0.0);
 
+  // this is only for sim
+  protected double positionSetpoint = 0.0;
```
SCool62, line 61: Again could be wrong but i thought the point of CTRE sim is that you use the control requests as if it were a real robot,

src/main/java/frc/robot/elevator/ElevatorIOCTRESim.java
spellingcat: i suspect it's because it reconverts it back with the sensor to mech ratio lol

spellingcat:
moved to ctre-sim-rabbit-hole

